PR #4, opened 2026-08-02
Add the MIT license

Maximilian-Nesslauer:
MIT, matching StarMap and the rest of the KSA modding ecosystem. Copyright line names you as the original creator, per the transfer condition.

Answering the license question from Discord: I would not use CC BY-NC for any of it. NC makes the code non-free and incompatible with almost everything, and Creative Commons themselves advise against CC licenses for software. Plain MIT keeps every option open.

Your call in the end, it is your project.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,22 @@
+MIT License
+
+Copyright (c) 2026 Jeremy Harrington (MrJeranimo) and Borea contributors
+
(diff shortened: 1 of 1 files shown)


MrJeranimo (approved):
I approve. Same license I've been using for most my work anyways.

averageksp (approved):

PR #21, opened 2026-08-04
Notify Discord on CI failures

Maximilian-Nesslauer:
Posts a message to the github-activity channel when a build run completes with a failure, using the `DISCORD_WEBHOOK_URL` repo secret. Successful runs stay silent; the repo webhook no longer sends any Actions events, so this workflow is the only CI noise the channel gets.

### DIFF
--- a/.github/workflows/notify-failure.yml
+++ b/.github/workflows/notify-failure.yml
@@ -1,0 +1,24 @@
+name: notify-failure
+
+on:
+  workflow_run:
+    workflows: [build]
+    types: [completed]
+
+permissions: {}
+
+jobs:
+  notify:
+    name: notify
+    if: github.event.workflow_run.conclusion == 'failure'
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


MrJeranimo:
Should we not add in release notifications or merges to main? or will that still be a lot?

Maximilian-Nesslauer:
> Should we not add in release notifications or merges to main? or will that still be a lot?

They are included. I just changed the settings in: https://github.com/KSAModding/Borea/settings/hooks/660025339

I changed from "Send me **everything**. " to the ones you can see listed now. Let me know if that sounds fine for you. We might want to allow more of these items such as " A star is created or deleted from a repository. " but those can be done in the webhook config and dont need to be done here.

The reason this PR exists is that you can only filter by the event **type** but not the result. E.g. you can only turn on/off "notify on both the check successes or failures" but you cant only activate/deactivate it for one type.

averageksp (approved):
LGTM

MrJeranimo (approved):

Issue #15, opened 2026-08-04
CI: test on Linux, Windows, MacOS, add coverage reporting


Maximilian-Nesslauer:
The current CI builds and tests on Ubuntu only. Borea calls itself cross-platform, so the CI should test cross-platform too.

## 1.

Run build and test on `ubuntu-latest`, `windows-latest`, and `macos-latest`.

## 2. Coverage via Coveralls

Collect coverage on the Ubuntu leg with https://coveralls.io/ 

Deliberately no hard coverage threshold. I just want to see the coverage but i dont want to force the addition of "stupid" tests just because the coverage report demands it.

## 3. Dependabot

update PRs for GitHub Actions versions and other packages.

## 4. Optional: format check

`dotnet format --verify-no-changes` as a CI step, plus an `.editorconfig`.

If this sounds right, I will open the PR.

averageksp:
Sounds good

MrJeranimo:
Is 2, Coveralls, in the repo yet? If so, how do I access it?

Maximilian-Nesslauer:
> Is 2, Coveralls, in the repo yet? If so, how do I access it?

No it still needs to be added :) I have not had  the time to do that yet

PR #23, opened 2026-08-05
Allowed Version, BuiltForGameVersion, ReleasedAt, and FileSizeBytes to be null in `ModMetadata`

### LINKED ISSUE #22: Allow certain properties of `ModMetadata` to be null/empty.
In preparation for [#15](https://github.com/KSAModding/content-manager-design/discussions/15) `ModMetdata` should be updated to allow certain fields that will not be in the Authored Metadata file from the indexer. As the authored file may be received without the versioned file (as intended) certain fields must be allowed to be null. 

The other option is for the ModMetadata to be split up into two different files, one fore the author file, and one for the versioned.

Later the Networking files must be updated as well.

MrJeranimo:
This PR may be too early as I realized we will have to restructure ModMetadata based on what passes in [#15](https://github.com/KSAModding/content-manager-design/discussions/15) no matter what.


Maximilian-Nesslauer (commented):
`ModMetadataMapper.ToDto` still calls `metadata.Version.ToString()`, but `Version` is a `ModVersion?` now, and `Nullable<T>.ToString()` returns the empty string when there is no value, not `null`.
So a metadata without a version is persisted as `""`, and `FromDto` then takes the `dto.Version is not null` branch and calls `ModVersion.Parse("")`, which throws `FormatException` because `TryParse` rejects whitespace.
Same for `BuiltForGameVersion`.
`metadata.Version?.ToString()` fixes it, and a round-trip test with the fields absent would have caught it, which is why CI is green here.

`fileSizeBytes` can still be wrong when it is present, so I would keep the check as `if (fileSizeBytes is < 0)` and keep the test.

**On the shape.**
I would go with the other option from #22, two types rather than one type with nullable halves.

The authored file and the generated file are different documents with different writers and different lifetimes, and four nullable fields do not encode which of the two you are holding, so every consumer has to guess and none of them can be checked by the compiler.

Splitting it makes "authored without a release" a valid state by construction instead of a state that has to be validated everywhere.

You are right that this restructuring follows from #15 either way, so I would not rush it in.

MrJeranimo:
I'm going to close this pull request since you pointed out we are going to have to refactor anyway.

Issue #31, opened 2026-08-08
Pack metadata: align ModPackMetadata with RFC 0031


Maximilian-Nesslauer:
Part 5 of the split in #25, alongside storage (#26), network (#27), and the resolver (#28), because packs are covered by none of them.

`ModPackMetadata` still carries the pre-RFC shape.
[RFC 0031](https://github.com/KSAModding/content-manager-design/blob/main/rfcs/0031-content-metadata-format.md) defines a pack as the shared authored core plus a small extension, so the class needs:

- the shared core: `spec_version`, `type`, `authors` as a list (currently a single string), required `abstract` with optional `description` (currently one required description), `license`, `tags`, `status` and `superseded_by`, and `links` with the required forums entry,
- game compatibility as `game_min` (which may be a month, a form `GameVersion` cannot represent) with optional `game_max` and `os`, replacing the single `BuiltForGameVersion`,
- the pack extension: `version`, `released_at`, optional `changelog`, the pinned `[[mods]]` list, and the `[[vehicles]]` and `[[saves]]` sections with the same entry shape,
- pack ids validated by the same id rules as mods (`ModIds`), since the namespace is global across content types.

A pack document is immutable once published, so the class can stay a plain snapshot: there is no generated half and no watcher involved.

Maximilian-Nesslauer:
issue reopened because it was closed but never (fully) implemented. It was accidentally swept up in the batch close of #29 , #25 , #26 

#26 really was implemented, by #33. #31 was not

Issue #30, opened 2026-08-08
ModVersion: SemVer 2.0.0 pre-release ordering and build metadata


Maximilian-Nesslauer:
`ModVersion.CompareTo` resolves pre-release precedence with `string.CompareOrdinal(PreRelease, other.PreRelease)`, which is not SemVer ordering: numeric identifiers must compare numerically.
`CompareOrdinal("beta.10", "beta.9") < 0`, so `1.0.0-beta.10` sorts below `1.0.0-beta.9`, and every bounds check built on `CompareTo` (dependency, any_of, and loader bounds) inherits the wrong answer.
[RFC 0031](https://github.com/KSAModding/content-manager-design/blob/main/rfcs/0031-content-metadata-format.md) pins the release version to SemVer 2.0.0 including its pre-release ordering, so this is a spec conformance bug, not a style choice.

Two parse holes feed the same comparator:

- `TryParse` rejects valid build metadata outright: `1.0.0+build` fails because the third part becomes `0+build`.
- For `1.0.0-beta+exp` the `+exp` is folded into `PreRelease`, so build metadata participates in precedence, which SemVer 2.0.0 forbids.

Fix:

- Strip `+build` in `TryParse` before splitting the core version, and exclude it from precedence (dropping it from the stored value is fine, the index normalizes versions at stamp time anyway).
- Compare pre-release labels identifier by identifier per SemVer 2.0.0: split on `.`, numeric identifiers compare numerically and rank below alphanumeric ones, and when all shared identifiers are equal the shorter list ranks lower.
- Tests for the ordering table in the SemVer spec, including `alpha < alpha.1 < alpha.beta < beta < beta.2 < beta.11 < rc.1 < (release)`.

Issue #28, opened 2026-08-06
Resolver: evaluate the five dependency kinds


Maximilian-Nesslauer:
Part 4 of the split in #25, blocked on the Core model PR.

Teach `ModDependencyResolver` the semantics of `ModDependencyKind` per RFC #31 (content-manager-design):

- `required` installs always, `recommends` is selected by default and deselectable, `suggests` is listed but not selected, `optional` is not installed by default.
- `conflict` blocks installing both within the bounds; no bounds means every version conflicts.
- Version bounds are inclusive `min`/`max` with an open upper end when absent.
- Tests per kind, including the no-bounds conflict case.

Issue #26, opened 2026-08-06
Storage: persist ModMetadata and ModVersionMetadata


Maximilian-Nesslauer:
Part 2 of the split in #25, blocked on the Core model PR.

Add DTOs and mappers for `ModMetadata` and `ModVersionMetadata`, one DTO plus one mapper per type, following the existing `TomlFileStore` pattern.

- Round-trip tests for both types, including every optional field absent.
- Migrate whatever `FileModStateRepository` and friends persisted under the old single-class shape, or document why nothing stored so far needs migrating.
- Absent optional fields must round-trip as absent, not as `""` or a default that later parses differently.

PR #32, opened 2026-08-08
Align the metadata classes with RFC 0031

Maximilian-Nesslauer:
The field-mapping checklist from #25, implemented on top of this branch and cross-checked against [RFC 0031](https://github.com/KSAModding/content-manager-design/blob/main/rfcs/0031-content-metadata-format.md).

**What changes**

- The missing release facts land: `download` (url, sha256, sizes, content type, mirrors), `install`, `install_size`, `version_scheme`, `yanked`, and the `listing` snapshot. The sha256 becomes a hex string; it was typed as the `SHA256` algorithm class.
- The required `type` field lands as `ContentType`, so mod-loader listings are representable, with the RFC's type rules enforced.
- Game bounds follow the stamped form, display string plus revision integer, validated as a pair and cross-checked against each other (#13).
- The loader is one nullable `LoaderRequirement`, `[releases]` a structured `ReleaseSource` with an authority, links a dictionary with forums required, and absent optional fields are null instead of `""`.
- Dependencies gain `source` and `any_of`, and the new `ModIds` helper applies the RFC id rules (validation and case-insensitive comparison) consistently, including in `Instance` and `ModListComparer`.
- `InstalledMod` records its release's stamped dependency list, and the resolver evaluates required entries on it, including the any_of cases. Full kind semantics stay #28.
- `VersionRange` and its tests are removed, since the format settled on inclusive min/max bounds (#13).

**Review notes**

Seven commits, each a reviewable unit. Intermediate commits do not build on their own (the base does not either); the head builds clean and Borea.Core.Tests passes with 232 tests, updated to the new model.
Storage, Network, and their tests stay for #26/#27, per the split in #25.

Deliberately not here: SemVer pre-release ordering in `ModVersion` (#30) and the pack metadata alignment (#31). `IModRepository` also has no path to `ModVersionMetadata` yet, which is worth deciding on #25 first.


PlazmaBoltz (dismissed):

src/Borea.Core/Mods/InstalledMod.cs
```diff
@@ -1,38 +1,56 @@
...
     public string ModId { get; }
     public ModVersion Version { get; }
     public InstallReason Reason { get; private set; }
     public DateTimeOffset InstalledAt { get; }
+
+    /// <summary>
+    /// The live authored listing. Its Dependencies are the authored entries;
+    /// the resolver acts on <see cref="Dependencies"/>, the stamped list, instead.
+    /// </summary>
     public ModMetadata Metadata { get; }
```
MrJeranimo: Should we not change this to `ModVersionMetadata` as an installed mod is a specific version and thus should link to the `ModVersionMetadata`.

src/Borea.Core/Mods/InstalledMod.cs
```diff
@@ -1,38 +1,56 @@
...
+
+    /// <summary>
+    /// The dependency list of the installed release, as its release file stamped it.
+    /// </summary>
+    public IReadOnlyList<ModDependency> Dependencies { get; }
```
MrJeranimo: If the above comment is accepted then we can remove this since the dependency list can be found in the `ModVersionMetadata`, thus removing the duplication.

MrJeranimo (dismissed):
For `InstalledMod` I think we should change the `ModMetadata` field to `ModVersionMetadata` since an `InstalledMod` is a specific version of a mod that is installed. This also removes the duplication of the Dependencies field from `InstalledMod` and you can just find it by going `InstalledMod.ModVersionMetadata.Dependencies`.

MrJeranimo (approved):

PlazmaBoltz (approved):

Issue #27, opened 2026-08-06
Network: map SpaceDock data onto the new metadata types


Maximilian-Nesslauer:
Part 3 of the split in #25, blocked on the Core model PR, independent of the storage work.

Rework `SpaceDockModRepository` and its mappers to produce the new types.

- SpaceDock cannot provide most authored fields (license list, dependencies, loader, forums link), so decide per field between a placeholder and an explicit unknown.
- A mod whose data does not parse must surface as an entry with unknown state instead of disappearing from the list, which is what `TryMapToMetadata` returning null does today.
- Keep the `Source` tag so index-backed and SpaceDock-backed entries stay distinguishable.

PR #33, opened 2026-08-08
Storage: persist the new metadata model

Maximilian-Nesslauer:
Implements #26 on top of #32: the storage layer for the new metadata model, one DTO plus one mapper per type, following the `TomlFileStore` pattern.

**What changes**

- DTOs and mappers for `ModMetadata` and `ModVersionMetadata` with their support types (dependencies including `any_of`, loader, release source, download, install, listing snapshot). Absent optional fields are nullable and round-trip as absent, guarded by tests on both the reloaded objects and the serialized text.
- `DateTimeOffset` is persisted as an ISO 8601 round-trip string through a converter in `TomlFileStore`, because the plain TOML datetime form drops fractional seconds; this also fixes the existing `InstalledAt`/`CreatedAt` precision loss.
- The enum vocabulary is written in the format's lowercase spelling through one explicit mapper with Unknown fallbacks for unrecognized values, an absent `status` reads as active and an absent `version_scheme` as semver, matching [RFC 0031](https://github.com/KSAModding/content-manager-design/blob/main/rfcs/0031-content-metadata-format.md)'s defaults.
- Stored ids compare case-insensitively everywhere (manifest state, both favorites stores, and the uninstaller, which now resolves the mod folder by comparison for case-sensitive filesystems).
- `GetAllAsync` skips an instance whose file cannot be read instead of failing the whole list, and a file without a spec version gets a precise "predates the metadata model" error.

**Migration**: nothing has shipped. The manifest, favorites, and active-instance files are unchanged; `instances.toml` did change shape, so a pre-branch file on a dev machine fails its direct load with the error above (and is skipped in the list) and should simply be deleted.

These DTOs mirror Borea's local persistence, not the RFC wire format; the index reader in #27 gets its own.
Borea.Storage builds clean and its suite passes with 104 tests. SemVer build metadata in `ModVersion` stays #30.

### DIFF
--- a/src/Borea.Storage/Mods/InstalledModDto.cs
+++ b/src/Borea.Storage/Mods/InstalledModDto.cs
@@ -1,17 +1,20 @@
 ﻿namespace Borea.Storage.Mods;
 
 /// <summary>
-/// Flat, TOML-serializable representation of an InstalledMod.
+/// TOML-serializable representation of an InstalledMod.
 /// </summary>
 public sealed class InstalledModDto
 {
     public string ModId { get; set; } = string.Empty;
     public string Version { get; set; } = string.Empty;
 
-    /// <summary>InstallReason as a string — "Manual", "ModPack", or "Dependency".</summary>
+    /// <summary>InstallReason as a string - "Manual", "ModPack", or "Dependency".</summary>
     public string Reason { get; set; } = string.Empty;
 
     public DateTimeOffset InstalledAt { get; set; }
     public string? Checksum { get; set; }
     public ModMetadataDto Metadata { get; set; } = new();
+
+    /// <summary>The dependency list of the installed release, as its release file stamped it.</summary>
+    public List<ModDependencyDto> Dependencies { get; set; } = new();
 }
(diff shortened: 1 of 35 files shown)


src/Borea.Storage/Mods/InstalledModDto.cs
```diff
@@ -1,17 +1,20 @@
...
     public string ModId { get; set; } = string.Empty;
     public string Version { get; set; } = string.Empty;
 
-    /// <summary>InstallReason as a string — "Manual", "ModPack", or "Dependency".</summary>
+    /// <summary>InstallReason as a string - "Manual", "ModPack", or "Dependency".</summary>
     public string Reason { get; set; } = string.Empty;
 
     public DateTimeOffset InstalledAt { get; set; }
     public string? Checksum { get; set; }
     public ModMetadataDto Metadata { get; set; } = new();
```
MrJeranimo: Will probably need to update this line to `ModVersionMetadata` if the changes in #32 gets approved.

MrJeranimo (changes requested):

src/Borea.Storage/Mods/InstalledModDto.cs
Maximilian-Nesslauer: Done in 8fea1d0: the DTO now embeds the release as ModVersionMetadataDto, and the separate dependency list is gone with it.

PlazmaBoltz (approved):

MrJeranimo (approved):

PlazmaBoltz (approved):

Issue #38, opened 2026-08-12
Read the installed game version from KSA.dll's version resource


Maximilian-Nesslauer:
Borea needs to read which game build is installed.

[RFC 0017](https://github.com/KSAModding/content-manager-design/blob/main/rfcs/0017-game-version-ordering-and-compatibility.md) gives an ordered chain: `KSA.dll` PE FileVersion, then PE ProductVersion, then the `build` field of the newest `Content/Versions/*.json`. This issue covers the first two; the `Content/Versions` fallback is a follow-up.

- [ ] `IInstalledGameVersionProvider` in `Borea.Core/Game`, implemented in `Borea.Storage/Game`, following the `ILatestVersionPing` and `IGamePathProvider` precedent of interface in Core, I/O outside it.
- [ ] Take the directory from `IGamePathProvider.GetGameDirectoryPath()`, which is nullable. Read `FileVersionInfo`, try `FileVersion` then `ProductVersion`. `GameVersion.TryParse` already accepts a leading `v`, a `-suffix` and a `+hash`, so ProductVersion parses unchanged.
- [ ] Return the parsed `GameVersion?` next to the raw string, mirroring `LatestVersionInfo`. RFC 0017 says an unparseable version "becomes Unknown rather than being discarded", so the string stays displayable.
- [ ] Three failure modes: the game path is null, `KSA.dll` is missing (`GetVersionInfo` throws), and the file has no version resource (`FileVersion` comes back null).
- [ ] Tests in `Borea.Storage.Tests`, where `TestGamePathProvider` already roots a temp game folder per test: path null, `KSA.dll` missing, present but unparseable, present and valid.

For the valid case add a tiny fixture class library under `tests/Borea.Storage.Tests/Fixtures/` with `<FileVersion>2026.8.3.5117</FileVersion>` and copy its output in as `KSA.dll`.

Do not build a read seam around `FileVersionInfo`.

PR #55, opened 2026-08-31
Evaluate a release against the installed game and the target platform

### LINKED ISSUE #39: Evaluate compatibility per RFC 0017 and the os list per RFC 0031
`ModVersionMetadata` carries `GameMinRevision`, `GameMaxRevision` and `Os`.

[RFC 0017](https://github.com/KSAModding/content-manager-design/blob/main/rfcs/0017-game-version-ordering-and-compatibility.md) defines four states:

| State | Condition |
|---|---|
| Unknown | No usable lower bound. |
| Incompatible | The installed revision is below the lower bound. |
| Compatible | The upper bound is absent, or the installed revision is at or below it. |
| Untested | The installed revision is above a stated upper bound. |

Incompatible is the only state that blocks.
This classifies only; what a client does with Untested and Unknown is a UI decision and not part of this issue.

**Shape**, in `Borea.Core`:

- [ ] `enum GameCompatibility { Compatible, Untested, Incompatible, Unknown }`
- [ ] `Evaluate(int? minRevision, int? maxRevision, GameVersion installed)`, plus an overload taking a `ModVersionMetadata`.
- [ ] Compare `installed.Revision` against the bound integers directly.

**On Unknown:** `ModVersionMetadata` requires `gameMin` and a non-nullable `GameMinRevision`, so the release overload can never return Unknown and there is no "no bounds at all" case for a release. That is expected, not a gap to fix here. Unknown exists for the listing side, where `SpaceDockModRepository` maps an unparseable game version to `GameMin = "unknown"`. The nullable parameter is what makes the state reachable.

**The os list**, from the same `[compatibility]` table. RFC 0031: "A platform outside the stated `os` list warns and lets the user proceed: the list states what the author knows works, not everything that can." Advisory, never blocking.

- The vocabulary is `windows`, `linux`, `macos`; the RFC adds "New values arrive by RFC".
- [ ] The check takes the target platform as a parameter and never reads the running OS, so it stays testable on the three-OS CI matrix.
- [ ] Absent means no known restriction. Treat an empty list the same way, and say in a comment that this is our choice: the RFC defines only absent, and the domain deliberately keeps empty distinct from absent.
- [ ] Compare case-insensitively as client leniency; the RFC states no comparison rule for os.
- [ ] An unrecognised value must not make the list unusable and must not be silently dropped: keep evaluating the recognised entries and report the unrecognised ones, so a caller can say the listing knows a platform this build does not.
- [ ] Give the `Unknown` member a doc comment saying it means no usable compatibility data, because `ModStatus.Unknown` and `ReleaseStatus.Unknown` already use the name for "a value this client version does not know".

Maximilian-Nesslauer:
Closes #39.

`Compatibility.Evaluate` answers the four states of RFC 0017, ordered by the revision alone, and `EvaluateOs` answers the os list of RFC 0031.

The installed version is nullable, so a caller that could not detect or parse it answers Unknown. That is the shape #38 hands back.

No caller acts on either yet.

### DIFF
--- a/src/Borea.Core/Game/Compatibility.cs
+++ b/src/Borea.Core/Game/Compatibility.cs
@@ -1,0 +1,37 @@
+﻿using Borea.Core.Mods;
+
+namespace Borea.Core.Game;
+
+/// <summary>
+/// Classifies content against the installed game.
+/// </summary>
+public static class Compatibility
+{
+    /// <summary>
+    /// The state the revision bounds put the installed game in (RFC 0017).
+    /// </summary>
+    public static GameCompatibility Evaluate(int? minRevision, int? maxRevision, GameVersion? installed)
+    {
+        if (installed is not { } version || minRevision is not { } min)
+            return GameCompatibility.Unknown;
+
+        if (version.Revision < min)
+            return GameCompatibility.Incompatible;
+
+        return maxRevision is { } max && version.Revision > max
+            ? GameCompatibility.Untested
+            : GameCompatibility.Compatible;
+    }
+
+    /// <summary>
+    /// The state the bounds of a stamped release put the installed game in.
+    /// A release always carries a lower bound.
+    /// </summary>
+    public static GameCompatibility Evaluate(ModVersionMetadata release, GameVersion? installed)
+    {
+        if (release is null)
+            throw new ArgumentNullException(nameof(release));
+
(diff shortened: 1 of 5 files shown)


MrJeranimo (approved):
Quality work

PR #56, opened 2026-08-31
Align the pack document with RFC 0031

### LINKED ISSUE #31: Pack metadata: align ModPackMetadata with RFC 0031

Maximilian-Nesslauer:
Closes #31.

`ModPackMetadata` predates RFC 0031 and was the last type left on the pre-format shape.

It carries the shared authored core plus the pack extension now.

`ModPackEntry` moves to its own file and its id becomes `ContentId`, since the same shape now carries `[[vehicles]]` and `[[saves]]` as well. It validates the id and compares it case-insensitively, so entry equality cannot disagree with what the document calls a repeated pin.

### DIFF
--- a/src/Borea.Core/Comparison/ModListComparer.cs
+++ b/src/Borea.Core/Comparison/ModListComparer.cs
@@ -19,7 +19,7 @@ public sealed class ModListComparer
         if (targetMods is null) throw new ArgumentNullException(nameof(targetMods));
 
         var currentById = currentMods.ToDictionary(m => m.ModId, ModIds.Comparer);
-        var targetIds = new HashSet<string>(targetMods.Select(m => m.ModId), ModIds.Comparer);
+        var targetIds = new HashSet<string>(targetMods.Select(m => m.ContentId), ModIds.Comparer);
 
         var toAdd = new List<ModPackEntry>();
         var toUpdate = new List<ModVersionChange>();
@@ -27,12 +27,12 @@ public sealed class ModListComparer
 
         foreach (var target in targetMods)
         {
-            if (currentById.TryGetValue(target.ModId, out var current))
+            if (currentById.TryGetValue(target.ContentId, out var current))
             {
                 if (current.Version.Equals(target.Version))
-                    unchanged.Add(target.ModId);
+                    unchanged.Add(target.ContentId);
                 else
-                    toUpdate.Add(new ModVersionChange(target.ModId, current.Version, target.Version));
+                    toUpdate.Add(new ModVersionChange(target.ContentId, current.Version, target.Version));
             }
             else
             {
--- a/src/Borea.Core/Comparison/ModListComparer.cs
+++ b/src/Borea.Core/Comparison/ModListComparer.cs
@@ -4,14 +4,18 @@ using Borea.Core.Mods;
 namespace Borea.Core.Comparison;
 
 /// <summary>
-/// Compares a set of currently installed mods against a target mod list.
+/// Compares the mods installed in an instance against the mods a pack pins.
 /// </summary>
 public sealed class ModListComparer
 {
-
     /// <summary>
-    /// Compares a list of InstalledMods against a list of ModPackEntries,
-    /// returning a ModListDiff that describes the differences.
+    /// Compares the installed mods against the pinned mods and returns the
+    /// <see cref="ModListDiff"/> between them.
+    /// <paramref name="targetMods"/> is the mods section of a pack document, so
+    /// <see cref="ModPackMetadata.Mods"/> and never its vehicles or saves. A pin carries
+    /// no content type, the section it sits in does, and an installed mod is the only
+    /// thing this comparison matches a pin against, so an entry from another section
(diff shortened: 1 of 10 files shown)


src/Borea.Core/Comparison/ModListComparer.cs
```diff
@@ -19,20 +19,20 @@ public ModListDiff Compare(IReadOnlyList<InstalledMod> currentMods, IReadOnlyLis
         if (targetMods is null) throw new ArgumentNullException(nameof(targetMods));
 
         var currentById = currentMods.ToDictionary(m => m.ModId, ModIds.Comparer);
-        var targetIds = new HashSet<string>(targetMods.Select(m => m.ModId), ModIds.Comparer);
+        var targetIds = new HashSet<string>(targetMods.Select(m => m.ContentId), ModIds.Comparer);
```
MrJeranimo, line 26: Will there not be extra Ids in this since the compare is only comparing `InstalledMod`s to `ModPackEntry`s where `ModPackEntry` can be anything in a ModPack? Or will the `targetMods` variable only contain the mods by convention?

src/Borea.Core/ModPacks/ModPackEntry.cs
```diff
@@ -0,0 +1,25 @@
...
+    public string ContentId { get; }
+
+    public ModVersion Version { get; }
+
+    public ModPackEntry(string contentId, ModVersion version)
+    {
+        ModIds.Validate(contentId, nameof(contentId));
+
+        ContentId = contentId;
+        Version = version;
```
MrJeranimo, line 26: **[minor]**
`contentId` gets validated but `version` does not.

src/Borea.Core/Comparison/ModListComparer.cs
Maximilian-Nesslauer: By convention, yes.

A pin carries no content type, because RFC 0031 keeps every reference to an id type-free and lets the section say what kind of content the id names.

The comparer matches pins against installed mods, so its input is the mods section of a pack, `pack.Mods`, never the vehicles or saves.

An entry from another section would come back as a mod to add, which is what you describe.

Vehicles and saves will get their own comparison once their types have install semantics but we dont have an RFC for that yet.

I did not give the entry a type of its own, since that would put a type back on a pin and duplicate the section it sits in.
Nothing calls the comparer yet, so I left the surface alone for now.

src/Borea.Core/ModPacks/ModPackEntry.cs
Maximilian-Nesslauer: There is nothing left for the entry to check.

`ModVersion` is a readonly record struct that validates its components and the pre-release label in its constructor and in `Parse`, so an entry cannot hold a malformed version.

The one default value that skips the constructor is `default`, and that is `0.0.0`, just because it is a legal SemVer version.
A whole default entry is rejected by the pack constructor through its null id.

I now added a doc comment on `Version` that says this and a test that pins `default` down as `0.0.0`.

Maximilian-Nesslauer:
I did push the code for the the replies above as code, plus a i added some stuff after a second look over the whole diff:

- `RejectRepeatedPins` names the constructor parameter (`mods`, `vehicles`, `saves`) in the exception, the way every other guard in Core does, instead of the property name.
- Every public member of `ModPackMetadata` carries the same one-line summary its twin on `ModMetadata` has.
- I also added Tests: a repeated pin between mods and vehicles, a forums link with authored casing, an unknown status that parses to `Unknown`, an empty `os` list that stays empty, an absent type that is refused, and the constructor assertions for name, authors, abstract, license and version. The `ModPackEntry` tests moved into their own file.

src/Borea.Core/ModPacks/ModPackEntry.cs
MrJeranimo: Ah my misunderstanding

MrJeranimo (approved):

PR #58, opened 2026-08-31
Replace the single StarMap path with a path per loader

### LINKED ISSUE #42: Replace StarMapDirectoryPath with per-loader paths
`BoreaSettings` names one loader: it carries `GameDirectoryPath` and `StarMapDirectoryPath`, mirrored by `IGamePathProvider.GetStarMapDirectoryPath`, `GamePathProvider`, `BoreaSettingsDto` and `BoreaSettingsMapper`.

`GetStarMapDirectoryPath` has no caller outside the tests and the test double, `StarMapDirectoryPath` is only round-tripped by the mapper, and `GamePathProvider` is never constructed in `Borea.App`. There are no call sites to follow and no behaviour to preserve.

Installing a loader per RFC 0035 also needs `[provides].launch`, `content-dir` and the `[provides.configure]` write, and `[provides]` is never stamped into a release file so it has to be re-read from the authored listing.

The change:

- [ ] `BoreaSettings`: replace `StarMapDirectoryPath` with a loader-id-to-path map, rebuilt through `ModIds.Comparer` the way `ModMetadata` rebuilds `Links`. Validate each key with `ModIds.Validate` and keep the existing whitespace rejection per path.
- [ ] `IGamePathProvider`: `GetLoaderDirectoryPath(string loaderId)` returning null when unknown, and `GamePathProvider` follows.
- [ ] `BoreaSettingsDto` takes a dictionary like `ModMetadataDto.Links` so Tomlyn writes it as a table, and the mapper maps both directions.
- [ ] Tests including a round-trip with two loaders, and one with TOML keys differing only in case, which collide under `ModIds.Comparer` since TOML keys are case-sensitive and ids are not.

Maximilian-Nesslauer:
Closes #42.

`BoreaSettings` carries a loader-id-to-path map now, and `IGamePathProvider.GetStarMapDirectoryPath` becomes `GetLoaderDirectoryPath(loaderId)`. Ids compare case-insensitively through `ModIds.Comparer` .

It touches `TestGamePathProvider` on the two lines #57 also touches, so whichever of the two merges second needs to resolve them by keeping both.

### DIFF
--- a/src/Borea.Core/Paths/IGamePathProvider.cs
+++ b/src/Borea.Core/Paths/IGamePathProvider.cs
@@ -65,12 +65,12 @@ public interface IGamePathProvider
     string? GetGameDirectoryPath();
 
     /// <summary>
-    /// Root directory of the StarMap installation. Used as the base
-    /// path for launching StarMap.
+    /// Root directory of one installed mod loader. Ids compare
+    /// case-insensitively, and no id at all throws.
     /// </summary>
     /// <returns><list type="bullet">
-    /// <item>null if path is unknown.</item>
+    /// <item>null if no path is known for that loader.</item>
     /// <item>string path if known.</item>
     /// </list></returns>
-    string? GetStarMapDirectoryPath();
+    string? GetLoaderDirectoryPath(string loaderId);
 }
--- a/src/Borea.Core/Settings/BoreaSettings.cs
+++ b/src/Borea.Core/Settings/BoreaSettings.cs
@@ -1,4 +1,6 @@
-﻿namespace Borea.Core.Settings;
+﻿using Borea.Core.Mods;
+
+namespace Borea.Core.Settings;
 
 /// <summary>
 /// Borea's own cross-platform configuration. No app level settings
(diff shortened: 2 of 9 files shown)


MrJeranimo (dismissed):
Clean code is life

Maximilian-Nesslauer:
> Clean code is life

Thank you. I re-requested the review because your "Fixed Merge Conflict" commit dismissed your previous approval.

MrJeranimo (dismissed):

MrJeranimo (dismissed):
Stacked PRs suckkkkkkk. It won't show me where the merge conflict issssssss!!!!!

Maximilian-Nesslauer:
> Stacked PRs suckkkkkkk. It won't show me where the merge conflict issssssss!!!!!

give me a sec, i am working on fixing this 👍 Thank you very much for reviewing it. I already have an idea

MrJeranimo:
I just merged the damn wrong change while rebasing ... shoot

MrJeranimo (approved):

PR #59, opened 2026-08-31
Carry the authored install and provides tables

### LINKED ISSUE #43: Carry the authored [install] and [provides] tables from RFC 0035
`ModMetadata` carries `InstallRootOverride`, a single string, which is RFC 0031's `[install].root` and nothing else.
[RFC 0035](https://github.com/KSAModding/content-manager-design/blob/main/rfcs/0035-content-install-descriptor.md) replaced that table with a superset and added `[provides]` on a `mod-loader`.

## Scope

- [ ] `[install]` becomes a type of its own, replacing `InstallRootOverride`: `root`, `target`, `path`, `manages`, `steps`, `uninstall`. It shares the anchor enum with #41 rather than defining a second one.
- [ ] `[provides]`, only on `mod-loader`: `launch`, `content-dir`, `content-path`. `[provides.configure]`: `file`, `format`, `game-path`, a closed enum with one member in `spec_version = 1`.
- [ ] `steps` and `uninstall` are prose no manager may parse (rule 7), and absent is not empty (rule 8), so neither may default to an empty array.
- [ ] RFC 0035's Errors table, minus its last row, which is stamp-time and stays with the index. Two rows cut against the rest of the model and are worth naming: an unrecognised anchor or format is **invalid rather than ignored**, because guessing would write files somewhere the author did not choose; and `manages` may not claim `manifest.toml`, which `ModManifest.Save` regenerates from the game's own list, destroying anything it did not write.
- [ ] Storage DTO and mapper follow.

Not in scope: resolving an anchor to a real path and performing an install, and whether `manages` binds a manager or only informs it ([content-manager-design#20](https://github.com/KSAModding/content-manager-design/discussions/20)).

Maximilian-Nesslauer:
Closes #43.

The StarMap listing in the live index already carries the whole `[install]` and `[provides]` tables and`InstallDescriptor` and `LoaderProvides` carry them now.

### DIFF
--- a/src/Borea.Core/ModLoaders/ConfigureFormat.cs
+++ b/src/Borea.Core/ModLoaders/ConfigureFormat.cs
@@ -1,0 +1,16 @@
+namespace Borea.Core.ModLoaders;
+
+/// <summary>
+/// The format of a loader's own configuration file.
+/// </summary>
+public enum ConfigureFormat
+{
+    Json = 0,
+
+    Toml = 1,
+
+    /// <summary>
+    /// The index rejects it at publish time, a client keeps the listing.
+    /// </summary>
+    Unknown = 2,
+}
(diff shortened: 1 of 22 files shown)


MrJeranimo (approved):
Almost ready to start parsing the index

Issue #44, opened 2026-08-14
Read the index snapshot as a mod source


Maximilian-Nesslauer:
## Implementation

The implementation is proposed in #110.
This issue remains open until the pull request reaches the default branch.

`IModRepository` is implemented by `SpaceDockModRepository` and `CompositeModRepository`, and neither knows the community index exists.
Everything the RFC alignment milestone lands is therefore a model with no consumer: the metadata classes match RFC 0031, and nothing ever builds one from a real index document.

The contract is [spec/snapshot.md](https://github.com/KSAModding/content-manager-design/blob/main/spec/snapshot.md): one document at `https://ksamodding.github.io/content-index-releases/v1/index.json`, carrying both halves of the index plus the game release list, so search, resolution and compatibility all happen locally and offline.
A listing entry keeps three voices apart and never mixes them: the author writes `authored`, tooling writes `releases`, the index writes `index_status`.

## Scope

- [ ] An index-backed `IModRepository`, interface in `Borea.Core` and HTTP in `Borea.Network`, following `ILatestVersionPing`: `HttpClient` and `TimeProvider` injected, overlapping calls sharing one request.
- [x] Poll with `If-None-Match`, so an unchanged index costs one 304. The `ETag` lives next to the cached copy.
- [x] The cached snapshot lives in `Borea.Storage` and survives a restart, since offline-first is the point. It is JSON, so it does not go through `TomlFileStore`.
- [ ] Entries carry the Borea-internal `Source` tag. Composite precedence between the index and SpaceDock is the follow-up #35 named, and the index should win.
- [ ] `GetLatestReleaseAsync` skips yanked releases and `GetReleaseAsync` returns one as-is, which the interface already promises and only an index source can honour.

## The failure modes, which are the interesting half

- **A `snapshot_version` above what Borea implements** The user gets "this client is too old for the index".
- **A `spec_version` above it is per document**: that entry renders in an unknown state, is never dropped, and its id still counts as taken. The constant to compare against is #40.
- **An `index_status.state` Borea does not know warns and removes nothing.** Only `delisted` removes, and it arrives as a tombstone carrying id and status alone, so "removed" stays distinguishable from "never listed".
- **`releases: []` is valid**, not an error: the listing passed its checks before its host had a release.

Not in scope: downloading and installing, any UI, and packs beyond parsing them (#31).

## Depends on

[...]

PR #60, opened 2026-09-01
Add an index fetcher that fetches the Content Index using ETag `If-None-Match`

MrJeranimo:
Adds a class `ContentIndexFetcher` which grabs the index from the specified URI using an ETag with the header `If-None-Match` and stores it as `index.json`. The ETag lives in a sidecar file `index.json.etag` (yes it is weird but it avoids a lot of Path Provier boiler plate) and is read/updated if it exists, created if it doesn't, and deleted if the URI returns no ETag to prevent a stale ETag and index. The index is downloaded to a temp file `index.json.tmp` while the actual download is happening over HTTP and then the temp file is copied to `index.json` once the download is complete.

To allow access to the index, `IGamePathProvider` now has a method `GetIndexPath()` which returns the path to `index.json`.

### DIFF
--- a/src/Borea.Network/Index/ContentIndexFetcher.cs
+++ b/src/Borea.Network/Index/ContentIndexFetcher.cs
@@ -1,0 +1,63 @@
+﻿using System.Net.Http.Headers;
+
+namespace Borea.Network.Index
+{
+    public sealed class ContentIndexFetcher : IContentIndexFetcher
+    {
+        private readonly HttpClient _client;
+        private readonly Uri _indexUri;
+
+        public ContentIndexFetcher(HttpClient client, Uri indexUri)
+        {
+            _client = client;
+            _indexUri = indexUri;
+        }
+
+        public async Task<ContentIndexFetchResult> FetchAsync(string destinationPath, CancellationToken ct = default)
+        {
+            string etagPath = destinationPath + ".etag";
+            string tempPath = destinationPath + ".tmp";
+
+            using var request = new HttpRequestMessage(HttpMethod.Get, _indexUri);
+
+            if (File.Exists(destinationPath) && File.Exists(etagPath))
+            {
+                string existingEtag = (await File.ReadAllTextAsync(etagPath, ct)).Trim();
+                if (!string.IsNullOrEmpty(existingEtag))
+                {
+                    request.Headers.IfNoneMatch.Add(EntityTagHeaderValue.Parse(existingEtag));
+                }
+            }
+
+            using var response = await _client.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, ct);
+
+            if (response.StatusCode == System.Net.HttpStatusCode.NotModified)
+            {
+                return ContentIndexFetchResult.NotModified;
+            }
+
+            response.EnsureSuccessStatusCode();
+
+            await using (var tempStream = File.Create(tempPath))
+            await using (var responseStream = await response.Content.ReadAsStreamAsync(ct))
+            {
+                await responseStream.CopyToAsync(tempStream, ct);
+            }
+
+            File.Move(tempPath, destinationPath, overwrite: true);
+
+            string? newEtag = response.Headers.ETag?.Tag;
+
+            if (newEtag is not null)
+            {
+                await File.WriteAllTextAsync(etagPath, newEtag, ct);
+            }
+            else if (File.Exists(etagPath))
+            {
+                File.Delete(etagPath);
+            }
+
+            return ContentIndexFetchResult.Downloaded;
+        }
+    }
+}
--- a/src/Borea.Network/Index/ContentIndexFetcher.cs
+++ b/src/Borea.Network/Index/ContentIndexFetcher.cs
@@ -138,7 +138,7 @@ public sealed class ContentIndexFetcher : IContentIndexFetcher
             try
             {
                 // Uses a unsigned integer to auto-rejct negative values
-                if(!versionElement.TryGetUInt32(out uint SnapshotVersion))
+                if (!versionElement.TryGetUInt32(out uint SnapshotVersion))
                 {
                     throw new HttpRequestException("The index has an invalid 'snapshot_version'.");
                 }
--- a/src/Borea.Network/Index/ContentIndexFetcher.cs
+++ b/src/Borea.Network/Index/ContentIndexFetcher.cs
@@ -19,8 +19,7 @@ public sealed class ContentIndexFetcher : IContentIndexFetcher
     /// Fetches the content index from the remote server and saves it to the specified destination path.
     /// Uses an ETag to determine if the index has changed since the last fetch.
     /// </summary>
-    /// <param name="destinationPath"></param>
-    /// <param name="ct"></param>
+    /// <param name="destinationPath">The full path to the file where the index will be saved. (ex.g C:/Users/username/%LocalAppData%/Borea/index.json)</param>
     /// <exception cref="HttpRequestException"></exception>
     public async Task<ContentIndexFetchResult> FetchAsync(string destinationPath, CancellationToken ct = default)
     {
@@ -62,14 +61,10 @@ public sealed class ContentIndexFetcher : IContentIndexFetcher
 
         response.EnsureSuccessStatusCode();
 
-        // Make sure index is JSON or plain text
-        if (response.Content.Headers.ContentType?.MediaType != "application/json" && response.Content.Headers.ContentType?.MediaType != "text/plain")
-        {
-            throw new HttpRequestException($"Expected content type 'application/json' or 'text/plain' but got '{response.Content.Headers.ContentType?.MediaType}'");
-        }
-
         // Read the content as a byte array to then write to disk.
         byte[] body = await response.Content.ReadAsByteArrayAsync(ct);
+
+        await BasicIndexFormatCheckAsync(body, response, ct);
 
         Directory.CreateDirectory(Path.GetDirectoryName(destinationPath)!);
 
@@ -98,5 +93,86 @@ public sealed class ContentIndexFetcher : IContentIndexFetcher
         }
 
         return ContentIndexFetchResult.Downloaded;
+    }
+
+    /// <summary>
+    /// Checks if the index is JSON or plain text, if it can be parsed as JSON, and if it has 4 required properties:
+    /// 'snapshot_version', 'listings', 'packs', and 'game_versions'. The actual values of the properties (excluding 'snapshot_version') are not
+    /// validated, only their existence is checked.
+    /// </summary>
+    /// <exception cref="HttpRequestException"></exception>
+    private static async Task BasicIndexFormatCheckAsync(byte[] body, HttpResponseMessage response, CancellationToken ct)
+    {
+        // Make sure index is JSON or plain text
+        if (response.Content.Headers.ContentType?.MediaType != "application/json" && response.Content.Headers.ContentType?.MediaType != "text/plain")
+        {
+            throw new HttpRequestException($"Expected content type 'application/json' or 'text/plain' but got '{response.Content.Headers.ContentType?.MediaType}'");
+        }
+
+        JsonDocument document;
+        try
+        {
+            document = await JsonDocument.ParseAsync(new MemoryStream(body), cancellationToken: ct);
+        }
+        catch (JsonException ex)
+        {
+            throw new HttpRequestException("Failed to parse index as JSON", ex);
+        }
+
+        using (document)
+        {
+            JsonElement root = document.RootElement;
+
+            if (root.ValueKind != JsonValueKind.Object)
+            {
+                throw new HttpRequestException("The index is not a JSON object.");
+            }
+
+            if (!root.TryGetProperty("snapshot_version", out var versionElement))
+            {
+                throw new HttpRequestException("The index has no usable 'snapshot_version'.");
+            }
+
+            int snapshotVersion;
+
+            try
+            {
+                if(!versionElement.TryGetInt32(out int SnapshotVersion))
+                {
+                    throw new HttpRequestException("The index has an invalid 'snapshot_version'.");
+                }
+                snapshotVersion = SnapshotVersion;
+            }
+            catch (InvalidOperationException ex)
+            {
+                throw new HttpRequestException("The index has an invalid 'snapshot_version'.", ex);
+            }
+
+            // Change 1 to SnapshotVersions.Highest once implemented
+            if (snapshotVersion > 1)
+            {
+                throw new HttpRequestException(                                     // Same here
+                    $"The index is snapshot version {snapshotVersion} and this build reads {1}.");
+            }
+
+            // Not checking for 'sources' since it is optional
+            // For these three properties, we don't care about the value since they are arrays or objects.
+            // Only checking for their existence for now. Reading the index client side can do a full validation.
+            // This can be changed later if a IndexValidator class or similar is implemented to validate the index fully.
+            if (!root.TryGetProperty("listings", out var listingsElement))
+            {
+                throw new HttpRequestException("The index has no usable 'listings'.");
+            }
+
+            if (!root.TryGetProperty("packs", out var packsElement))
+            {
+                throw new HttpRequestException("The index has no usable 'packs'.");
+            }
+
+            if (!root.TryGetProperty("game_versions", out var gameVersionsElement))
+            {
+                throw new HttpRequestException("The index has no usable 'game_versions'.");
+            }
+        }
     }
 }
--- a/src/Borea.Network/Index/ContentIndexFetcher.cs
+++ b/src/Borea.Network/Index/ContentIndexFetcher.cs
@@ -56,22 +56,27 @@ public sealed class ContentIndexFetcher : IContentIndexFetcher
         }
         else if (response.StatusCode == System.Net.HttpStatusCode.NotModified)
         {
-            // I am not sure what to throw here but the caller should know that the response was not what was expected.
+            // This is to prevent a situation where we should've downloaded the index, yet didn't due to a incorrect 304 response.
             throw new HttpRequestException($"ETag does not exist yet request returned 304 Not Modified: {response.ReasonPhrase}");
         }
 
         response.EnsureSuccessStatusCode();
 
+        // Make sure index is JSON or plain text
+        if (response.Content.Headers.ContentType?.MediaType != "application/json" && response.Content.Headers.ContentType?.MediaType != "text/plain")
+        {
+            throw new HttpRequestException($"Expected content type 'application/json' or 'text/plain' but got '{response.Content.Headers.ContentType?.MediaType}'");
+        }
+
         // Read the content as a byte array to verify the JSON structure before writing to disk.
         byte[] body = await response.Content.ReadAsByteArrayAsync(ct);
 
-        VerifyIndexBasics(response.Content, body);
+        Directory.CreateDirectory(Path.GetDirectoryName(destinationPath)!);
 
         try
         {
             // Write the index to disk in a temp file then move to real file.
             // This is to avoid leaving a corrupted file if the process is interrupted.
-            Directory.CreateDirectory(Path.GetDirectoryName(destinationPath)!);
             await File.WriteAllBytesAsync(tempPath, body, ct);
             // Metadata change that only changes the file's name
             File.Move(tempPath, destinationPath, overwrite: true);
@@ -93,49 +98,5 @@ public sealed class ContentIndexFetcher : IContentIndexFetcher
         }
 
         return ContentIndexFetchResult.Downloaded;
-    }
-
-    /// <summary>
-    /// Verifies that the content index has the basic required fields and is valid JSON.
-    /// Only checks that "snapshot_version" and "sources" field exists.
-    /// </summary>
-    /// <returns>No return, only Exceptions</returns>
-    /// <exception cref="HttpRequestException"></exception>
-    private static void VerifyIndexBasics(HttpContent content, byte[] body)
-    {
-        if (content.Headers.ContentType?.MediaType != "application/json")
-        {
-            throw new HttpRequestException($"Expected content type 'application/json' but got '{content.Headers.ContentType?.MediaType}'");
-        }
-
-        JsonDocument document;
-        try
-        {
-            document = JsonDocument.Parse(body);
-        }
-        catch (JsonException ex)
-        {
-            throw new HttpRequestException("Failed to parse JSON content", ex);
-        }
-
-        using (document)
-        {
-            JsonElement root = document.RootElement;
-
-            if (root.ValueKind != JsonValueKind.Object)
-            {
-                throw new HttpRequestException("Expected JSON object as root element");
-            }
-
-            if (!root.TryGetProperty("snapshot_version", out _))
-            {
-                throw new HttpRequestException("Content index is missing required field 'snapshot_version'.");
-            }
-
-            if (!root.TryGetProperty("sources", out _))
-            {
-                throw new HttpRequestException("Content index is missing required field 'sources'.");
-            }
-        }
     }
 }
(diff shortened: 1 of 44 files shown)


src/Borea.Network/Index/ContentIndexFetcher.cs
```diff
@@ -0,0 +1,63 @@
...
+                return ContentIndexFetchResult.NotModified;
+            }
+
+            response.EnsureSuccessStatusCode();
+
+            await using (var tempStream = File.Create(tempPath))
+            await using (var responseStream = await response.Content.ReadAsStreamAsync(ct))
+            {
+                await responseStream.CopyToAsync(tempStream, ct);
+            }
```
Maximilian-Nesslauer: **[must fix]**

**1. The destination folder might never be created.**

`File.Create` cannot make a file inside a folder that does not exist, and `GetIndexPath()` points at `%LocalAppData%\Borea`, which might not be there on a fresh install until something else writes to it first.

Every other writer in the repo creates its directory first: `TomlFileStore` line 46, `SpaceDockModDownloader` lines 95 and 143.
I would add One line above the `File.Create`:

```csharp
Directory.CreateDirectory(Path.GetDirectoryName(destinationPath)!);
```

**2. The temp file is not cleaned up when the download fails part way.**
There is no `try`/`finally`, so a dropped connection or a cancellation leaves `index.json.tmp` behind:

```text
C) connection drops mid-body -> .tmp left behind (0 bytes); cached copy intact
```

`SpaceDockModDownloader.DownloadAsync` already wraps its temp file in a `finally` and deletes it, so the pattern is in the repo.

src/Borea.Network/Index/ContentIndexFetcher.cs
```diff
@@ -0,0 +1,63 @@
...
+            string tempPath = destinationPath + ".tmp";
+
+            using var request = new HttpRequestMessage(HttpMethod.Get, _indexUri);
+
+            if (File.Exists(destinationPath) && File.Exists(etagPath))
+            {
+                string existingEtag = (await File.ReadAllTextAsync(etagPath, ct)).Trim();
+                if (!string.IsNullOrEmpty(existingEtag))
+                {
+                    request.Headers.IfNoneMatch.Add(EntityTagHeaderValue.Parse(existingEtag));
```
Maximilian-Nesslauer: **[must fix] A damaged sidecar makes every later fetch throw**

`EntityTagHeaderValue.Parse` throws on anything that is not a valid entity tag, and the guard above only rejects an empty string.
The throw happens before the request is sent, so the one request that would rewrite the sidecar can never run.
The user has to delete the file by hand.

```text
B) corrupt / truncated sidecar -> THREW FormatException: The format of value '"abc' is invalid.
```

`TryParse` and falling through to an unconditional fetch is the same number of lines and self-heals:

```csharp
if (EntityTagHeaderValue.TryParse(existingEtag, out var tag))
    request.Headers.IfNoneMatch.Add(tag);
```

src/Borea.Network/Index/ContentIndexFetcher.cs
```diff
@@ -0,0 +1,63 @@
...
+
+            await using (var tempStream = File.Create(tempPath))
+            await using (var responseStream = await response.Content.ReadAsStreamAsync(ct))
+            {
+                await responseStream.CopyToAsync(tempStream, ct);
+            }
+
+            File.Move(tempPath, destinationPath, overwrite: true);
+
+            string? newEtag = response.Headers.ETag?.Tag;
```
Maximilian-Nesslauer: **[must fix] `.Tag` drops the weak marker, so a weak ETag does not round-trip.**

`EntityTagHeaderValue.Tag` returns the quoted part only; the weakness lives in the separate `IsWeak` property, and only `ToString()` puts the `W/` back.
So a server answering `ETag: W/"v2"` gets `"v2"` written to the sidecar and `"v2"` sent back as a strong validator:

```text
D) server sent  : W/"v2"
   sidecar holds: "v2"
   next request sends: "v2"
```

RFC 9110 requires weak comparison for `If-None-Match`, so a correct origin still answers 304 and nothing breaks today.
The risk is a proxy or CDN that does a plain byte compare: it answers 200 and Borea re-downloads the whole index on every launch, which removes the benefit this PR exists for.
The fetch URL is configuration per RFC 0033, so we do not control what sits in front of it forever.

```csharp
string? newEtag = response.Headers.ETag?.ToString();
```

A test with a weak tag would pin it.

src/Borea.Network/Index/ContentIndexFetcher.cs
```diff
@@ -0,0 +1,63 @@
...
+
+            response.EnsureSuccessStatusCode();
+
+            await using (var tempStream = File.Create(tempPath))
+            await using (var responseStream = await response.Content.ReadAsStreamAsync(ct))
+            {
+                await responseStream.CopyToAsync(tempStream, ct);
+            }
+
+            File.Move(tempPath, destinationPath, overwrite: true);
```
Maximilian-Nesslauer, line 77: I think that the rename is the right pattern, and same-folder means it is atomic.

**[should fix, but it may belong to #44 rather than to this PR. Your choice if you want to include it]**

`index.json` is Borea's only copy of the index, and it is what lets Borea work offline.
This line replaces it with whatever came back, and the only test before that is "was the status 200?".
Nothing has looked at the bytes yet.

### i believe that the following scenario is possible

You are in a hotel. The laptop is on the WiFi, but you have not signed in on the portal page yet.
In that state the router answers *every* HTTP request with 200 and an HTML sign-in page.

```text
E) 200 that is not the index -> cache now holds <html>Sign in to WiFi</html>
```

The portal sends no ETag either, so line 57 deletes the sidecar as well.

The next real download repairs it, but the damage still lands exactly in the situation where the offline cache would be needed.

`spec/snapshot.md` says a client that reads a `snapshot_version` higher than it implements "refuses the whole document and **keeps the copy it already has**", so the user keeps working on the old index while they update Borea.

Once this line has overwritten it, there is no copy left to keep, so no layer above can honour that rule.

src/Borea.Network/Index/ContentIndexFetcher.cs
```diff
@@ -0,0 +1,63 @@
...
+                    request.Headers.IfNoneMatch.Add(EntityTagHeaderValue.Parse(existingEtag));
+                }
+            }
+
+            using var response = await _client.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, ct);
+
+            if (response.StatusCode == System.Net.HttpStatusCode.NotModified)
+            {
+                return ContentIndexFetchResult.NotModified;
+            }
```
Maximilian-Nesslauer: **[minor]** A 304 is accepted even when no `If-None-Match` was sent.
A misbehaving proxy could then make `FetchAsync` return `NotModified` while `index.json` does not exist at all, and the caller would read a missing file.

Gating this on whether a validator was actually added would make sense.

src/Borea.Network/Index/ContentIndexFetcher.cs
```diff
@@ -0,0 +1,63 @@
...
+    public sealed class ContentIndexFetcher : IContentIndexFetcher
+    {
+        private readonly HttpClient _client;
+        private readonly Uri _indexUri;
+
+        public ContentIndexFetcher(HttpClient client, Uri indexUri)
+        {
+            _client = client;
+            _indexUri = indexUri;
+        }
```
Maximilian-Nesslauer: **[minor]** No argument validation, while the rest of the codebase does it everywhere: `LatestVersionPing`, `CompositeModRepository`, `FileInstanceRepository`, `FileModUninstaller` all throw `ArgumentNullException`, and `SpaceDockModDownloader.DownloadAsync` also rejects a null or whitespace destination path.

src/Borea.Network/Index/IContentIndexFetcher.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Borea.Network.Index
+{
+    public interface IContentIndexFetcher
```
Maximilian-Nesslauer: **[minor] Should this interface live in `Borea.Core`?**

Every other abstraction in Borea does: `ILatestVersionPing`, `IModRepository`, `IGamePathProvider`, `IModDownloader`, `IModUninstaller`, `IModStateRepository`, `IInstanceRepository`.
#44 asks for the same split for the index work, "interface in `Borea.Core` and HTTP in `Borea.Network`, following `ILatestVersionPing`".
As it stands, `Borea.App` and the planned in-game mod have to reference `Borea.Network` to depend on the abstraction, which is the thing that split avoids.

src/Borea.Network/Index/IContentIndexFetcher.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Borea.Network.Index
```
Maximilian-Nesslauer: **[minor]** These three new files use block-scoped namespaces.
On `src/` the count is 93 file-scoped against 5 block-scoped, and 3 of those 5 are this PR.
Not enforced by `.editorconfig`, so the format check stays green either way, it just stands out.

tests/Borea.Network.Tests/Index/ContentIndexFetcherTests.cs
```diff
@@ -0,0 +1,151 @@
...
+            await File.WriteAllTextAsync(_etagPath, "\"abc123\"");
+
+            var client = FakeHttpMessageHandler.BuildClient(
+                _ => new HttpResponseMessage(HttpStatusCode.InternalServerError),
+                out _);
+
+            await Assert.ThrowsAsync<HttpRequestException>(() => CreateFetcher(client).FetchAsync(_destinationPath));
+
+            Assert.Equal("""{ "mods": [] }""", await File.ReadAllTextAsync(_destinationPath));
+            Assert.False(File.Exists(_destinationPath + ".tmp"));
```
Maximilian-Nesslauer, line 219: **[should fix] This assertion passes without testing anything.**

`EnsureSuccessStatusCode()` throws on line 39 of `ContentIndexFetcher`, before `File.Create(tempPath)` on line 41, so on a 500 the temp file is never created in the first place.

The cases that would earn their keep here:

- the body fails part way through the download,
- the destination folder does not exist,
- the sidecar holds something that is not a valid entity tag,
- the server sends a weak ETag.

Maximilian-Nesslauer (changes requested):

Maximilian-Nesslauer:
Thanks, i think that this is good work and the sidecar should be the right call since you asked in discord.
It is what #44 already asks for with: "Poll with `If-None-Match`, so an unchanged index costs one 304. The `ETag` lives next to the cached copy."

I particularly like these 2 things:

- The download goes to `index.json.tmp` and is then renamed. Both paths sit in one folder, so the rename is atomic and a failed download can never leave a half-written `index.json`.
- The ETag is written *after* the index is in place. That is the safe order. The other order can leave Borea holding a new index while it believes the old one is current.

I built the branch and ran the fetcher against a fake server to check the failure paths.
Three of those paths need a fix, and I left them as inline comments:

1. A fresh install throws `DirectoryNotFoundException`, because `%LocalAppData%\Borea` might not exist yet.
2. A damaged `.etag` file throws `FormatException` on every later fetch, and the fetch that would repair it can never run.
3. A weak ETag (`W/"..."`) loses its `W/` prefix, so the caching this PR exists for can silently stop working behind a proxy.

The rest of the inline comments are optional.

I also test-merged this branch against the six other open branches with `git merge-tree`, including #58, which reworks the same `IGamePathProvider`, and all six merge clean.

src/Borea.Network/Index/ContentIndexFetcher.cs
MrJeranimo: I'll look into implementing it and see where it might go.

src/Borea.Network/Index/ContentIndexFetcher.cs
MrJeranimo: > `File.Create` cannot make a file inside a folder that does not exist, and `GetIndexPath()` points at `%LocalAppData%\Borea`, which might not be there on a fresh install until something else writes to it first.

I originally made this with the thought of Borea.App will make sure the directory exists, but I now also realize that a corrupted directory may cause an unfixable error, so I will definitely be fixing this and as many of the other changes as I can.

Maximilian-Nesslauer: > I originally made this with the thought of Borea.App will make sure the directory exists, but I now also realize that a corrupted directory may cause an unfixable error, so I will definitely be fixing this and as many of the other changes as I can.

Maybe i am just too cautious / overengineering things but it could not hurt handling that edge case :)

src/Borea.Network/Index/ContentIndexFetcher.cs
MrJeranimo: I implemented a super basic response body check. It checks that the body is JSON, it has an object as the root, and the root has two properties `snapshot_version` and `sources`. It doesn't check for any other specific properties or values as those might/will change, so I only check the two that are almost never going to change are there. 

The actual "Does the index have a valid format" check can be done in #44. This is only for checking is it JSON and does it somewhat look correct.

Let me know if we should change any part of this

src/Borea.Network/Index/ContentIndexFetcher.cs
```diff
@@ -0,0 +1,141 @@
...
+
+            if (!root.TryGetProperty("snapshot_version", out _))
+            {
+                throw new HttpRequestException("Content index is missing required field 'snapshot_version'.");
+            }
+
+            if (!root.TryGetProperty("sources", out _))
+            {
+                throw new HttpRequestException("Content index is missing required field 'sources'.");
+            }
```
Maximilian-Nesslauer: **[must fix] `sources` is optional, so this rejects valid snapshots.**

i believe that for presence checks , `listings`, `packs` and `game_versions` would make more sense.

My suggestion would be to just drop this block. 
Schema conformance belongs to #44.

src/Borea.Network/Index/ContentIndexFetcher.cs
```diff
@@ -0,0 +1,141 @@
...
+
+            if (root.ValueKind != JsonValueKind.Object)
+            {
+                throw new HttpRequestException("Expected JSON object as root element");
+            }
+
+            if (!root.TryGetProperty("snapshot_version", out _))
+            {
+                throw new HttpRequestException("Content index is missing required field 'snapshot_version'.");
+            }
```
Maximilian-Nesslauer: **`snapshot_version` is found but never compared**

`spec/snapshot.md` says that a client that reads a `snapshot_version` higher than it implements "refuses the whole document and **keeps the copy it already has**", and the user gets "this client is too old for the index".

If a `snapshot_version: 2` document is committed over `index.json` first, the copy to keep is already gone, so no layer above can honour that rule.

Borea has no constant for it yet.

I believe that we could leave this constant to #44, but maybe add a comment here so that we dont forget.

src/Borea.Network/Index/ContentIndexFetcher.cs
```diff
@@ -0,0 +1,141 @@
...
+    /// Only checks that "snapshot_version" and "sources" field exists.
+    /// </summary>
+    /// <returns>No return, only Exceptions</returns>
+    /// <exception cref="HttpRequestException"></exception>
+    private static void VerifyIndexBasics(HttpContent content, byte[] body)
+    {
+        if (content.Headers.ContentType?.MediaType != "application/json")
+        {
+            throw new HttpRequestException($"Expected content type 'application/json' but got '{content.Headers.ContentType?.MediaType}'");
+        }
```
Maximilian-Nesslauer: **[should fix] This rejects an address RFC 0033 names as valid.**

RFC 0033, on the fetch contract: "GitHub Pages and raw.githubusercontent.com serve strong ETags and 304s."

```text
ksamodding.github.io          -> Content-Type: application/json; charset=utf-8
raw.githubusercontent.com     -> Content-Type: text/plain; charset=utf-8
```

```text
text/plain -> THREW HttpRequestException: Expected content type 'application/json'
```

I would accept `text/plain` next to `application/json`.

src/Borea.Network/Index/ContentIndexFetcher.cs
```diff
@@ -0,0 +1,141 @@
...
+            // This is to avoid leaving a corrupted file if the process is interrupted.
+            Directory.CreateDirectory(Path.GetDirectoryName(destinationPath)!);
+            await File.WriteAllBytesAsync(tempPath, body, ct);
+            // Metadata change that only changes the file's name
+            File.Move(tempPath, destinationPath, overwrite: true);
+        }
+        finally
+        {
+            File.Delete(tempPath);
+        }
```
Maximilian-Nesslauer, line 82: **[should fix] The `finally` can hide the error.**

`File.Delete` does not throw when the file is missing, but it does throw `DirectoryNotFoundException` when the directory is missing. So if `Directory.CreateDirectory` on line 74 fails, the `finally` throws over the top of the original exception and the caller sees the wrong cause:

```text
folder cannot be created -> THREW DirectoryNotFoundException  (from File.Delete in the finally)
```

The real failure was `Directory.CreateDirectory`, and its exception is lost.

Either move line 74 above the `try`, since creating the directory is a precondition rather than something the cleanup has to undo, or guard the delete with `if (File.Exists(tempPath))`.

src/Borea.Network/Index/ContentIndexFetcher.cs
```diff
@@ -0,0 +1,141 @@
...
+
+        if (etagExists && response.StatusCode == System.Net.HttpStatusCode.NotModified)
+        {
+            return ContentIndexFetchResult.NotModified;
+        }
+        else if (response.StatusCode == System.Net.HttpStatusCode.NotModified)
+        {
+            // I am not sure what to throw here but the caller should know that the response was not what was expected.
+            throw new HttpRequestException($"ETag does not exist yet request returned 304 Not Modified: {response.ReasonPhrase}");
+        }
```
Maximilian-Nesslauer, line 60: **[minor]** regarding your comment: throwing here is right in my opinion.

`HttpRequestException` is fair, since a 304 that nobody asked for is a protocol answer that does not fit the request. I would just state that instead.

so i would just change the comment

Maximilian-Nesslauer (dismissed):
Thank you, the commits look good and all of my earlier points check out :)

Regarding your https://github.com/KSAModding/Borea/pull/60#discussion_r3910507003 comment: I believe that there is one thing to change before this merges because `sources` is optional in the spec, not required.

`spec/snapshot.md` lists it as `no` in the required column, and the builder in `content-index-releases` omits it on purpose whenever it cannot prove the provenance, "so a local build does not claim a provenance it does not have".

So rejecting a document without it turns away valid snapshots, including the fixture #44 wants us to build from `examples/`.

I wrote more details in the inline comments.

tests/Borea.Network.Tests/Index/ContentIndexFetcherTests.cs
MrJeranimo: I'm not fully sure what you mean here. Are you saying this test doesn't actually do anything since the Exception that is thrown is thrown because of a different condition and not the one being tested?

src/Borea.Network/Index/ContentIndexFetcher.cs
```diff
@@ -0,0 +1,102 @@
...
+        response.EnsureSuccessStatusCode();
+
+        // Make sure index is JSON or plain text
+        if (response.Content.Headers.ContentType?.MediaType != "application/json" && response.Content.Headers.ContentType?.MediaType != "text/plain")
+        {
+            throw new HttpRequestException($"Expected content type 'application/json' or 'text/plain' but got '{response.Content.Headers.ContentType?.MediaType}'");
+        }
+
+        // Read the content as a byte array to then write to disk.
+        byte[] body = await response.Content.ReadAsByteArrayAsync(ct);
```
Maximilian-Nesslauer, line 65: **[should fix] Bring the body check back, without `sources`.**

The three checks that already were in `VerifyIndexBasics` and should return: it parses as JSON, the root is an object, and `snapshot_version` is present. Only the `sources` check was wrong, because the spec has it as optional.

We could also compare `snapshot_version` instead of only finding it. `spec/snapshot.md` says a client reading a version higher than it implements "refuses the whole document and **keeps the copy it already has**". Once `File.Move` has run, there is no copy left to keep, so this is the last point where that rule can still be honoured.

Concretely, between reading the body and writing it:

```csharp
JsonDocument document;
try
{
    document = JsonDocument.Parse(body);
}
catch (JsonException ex)
{
    throw new HttpRequestException("The index is not valid JSON.", ex);
}

using (document)
{
    JsonElement root = document.RootElement;

    if (root.ValueKind != JsonValueKind.Object)
        throw new HttpRequestException("The index is not a JSON object.");

    if (!root.TryGetProperty("snapshot_version", out var versionElement) ||
        !versionElement.TryGetInt32(out int snapshotVersion))
    {
        throw new HttpRequestException("The index has no usable 'snapshot_version'.");
    }

    if (snapshotVersion > SnapshotVersions.Highest)
    {
        throw new HttpRequestException(
            $"The index is snapshot version {snapshotVersion} and this build reads {SnapshotVersions.Highest}.");
    }
}
```

`SnapshotVersions` does not exist yet.

```csharp
namespace Borea.Core.Index;

/// <summary>
/// The snapshot artifact format version, defined by spec/snapshot.md.
/// </summary>
public static class SnapshotVersions
{
    /// <summary>The highest snapshot version this build reads.</summary>
    public const int Highest = 1;

    /// <summary>Whether the snapshot is from a newer format than this build implements.</summary>
    public static bool IsAboveHighest(int snapshotVersion) => snapshotVersion > Highest;
}
```

I would also add some of the tests back you already wrote once: a truncated body, an array root, and a `snapshot_version` above the ceiling.

tests/Borea.Network.Tests/Index/ContentIndexFetcherTests.cs
Maximilian-Nesslauer: > Are you saying this test doesn't actually do anything since the Exception that is thrown is thrown because of a different condition and not the one being tested?

Yes, exactly that. The test asserts that no `.tmp` is left behind, but a 500 makes `EnsureSuccessStatusCode()` throw before the temp file is ever created, so the assertion is true. It would pass with the `finally` deleted.

***I believe that you have already fixed it.*** `FetchAsync_DownloadInterruptedMidStream_ThrowsAndLeavesCacheUntouched` fails part way through the body, which is the case where a temp file actually exists, so that one does exercise the cleanup. There should be nothing left to do here.

Maximilian-Nesslauer (commented):
Thank you very much for applying the changes. **This is fine work** :)

`text/plain` is accepted, `Directory.CreateDirectory` moved above the `try` so a failure there no longer hides behind the cleanup, and the throw on an unasked-for 304 now says why it is there. You even added better tests.

There is one thing that i believe is **_my_** (communication) fault.

I wrote "My suggestion: drop this block" on the `sources` check, and the sentence after it said "Valid JSON, an object root and `snapshot_version` are already enough". I meant drop the four lines the comment was attached to. You read it as the whole method, and now there is no body check at all.

`sources` was the only part that had to go and the fetcher accepts now:

| body | result | cached copy kept |
|---|---|---|
| the real index, `application/json` | accepted | no |
| the real index via raw, `text/plain` | accepted | no |
| captive portal HTML, `text/html` | rejected | yes |
| captive portal page as `text/plain` | accepted | **no** |
| proxy error page as `text/plain` | accepted | **no** |
| truncated JSON | accepted | **no** |
| JSON array root | accepted | **no** |
| `snapshot_version: 999` | accepted | **no** |
| empty body | accepted | **no** |

The media type is the only guard left, and since `text/plain` has to be accepted for `raw.githubusercontent.com`, it no longer separates the index from an error page.

Sry for the confusion. I wrote an inline comment for what i would do. It is your call whether the version ceiling goes in here or in #44.

Maximilian-Nesslauer (approved):
Thank you very much for being so active in this :) You work is outstanding.

Issue #52, opened 2026-08-29
Adopt a mod folder Borea did not install


Maximilian-Nesslauer:
## Implementation

The implementation is proposed in #113.
This issue remains open until the pull request reaches the default branch.

## Where this stands today

Borea knows about a mod only if Borea installed it.
The game does not work that way: `ModLibrary.AddMods` adds a manifest entry for any subfolder that contains a `mod.toml` and is not listed yet, creates it disabled, and logs it as an orphaned local mod.

So a folder a user dropped in is a working mod on the next launch, and Borea shows an instance that is not the instance the game runs.

## What it should be instead

| Today | Should be |
|---|---|
| Foreign folders are invisible to Borea. | The instance mods folder is scanned, and every folder with no `InstalledMod` behind it is listed. |
| An imported archive is nothing but bytes. | Its digest is looked up in the snapshot. Every stamped release carries `download.sha256` and RFC 0031 accepts any bytes matching it, so a match adopts the folder as that exact release, with its metadata, its bounds and its update path. |
| No match means no entry at all. | No match means the folder is kept with compatibility Unknown, which RFC 0017 already defines as warn and let the user proceed. No version can be recovered by inspection, because nothing on disk carries one. |
| A local mod takes no part in dependency resolution. | Its `[[StarMap.ModDependencies]]` are read from its own `mod.toml`, so it resolves as ids without version bounds. That is the same derived baseline the index watcher uses, so the two cannot disagree. |
| Uninstall has no rule about foreign folders. | A folder Borea did not install is never deleted. |

## Tests

An archive whose hash matches a listed release, one that does not, a folder without a `mod.toml`, and a folder matching nothing in the index.

PR #62, opened 2026-09-01
Follow the game's rules for manifest.toml

### LINKED ISSUE #53: Follow the game's rules for manifest.toml
## Where this stands today

`FileModStateRepository` reads `manifest.toml` into a `ManifestDto`, changes it and writes the whole file back.

## What it should be instead

| Today | Should be |
|---|---|
| `SetActiveAsync` appends a new entry at the end, so it decides load order as a side effect and says nothing about it. | Entry order is preserved, and reordering is an explicit operation. `ModLibrary.PrepareAll` walks the manifest in order and loads each enabled mod as it goes, and `ModManifest.IsChanged` treats a changed index as a change. |
| `SetActiveAsync` flips an existing entry to enabled. | `enabled = true` is written only for entries Borea creates, and never flipped on one it did not, because a mod somebody disabled in the game has to stay disabled through a Borea operation. |
| An entry can be written before its folder exists. | An entry is written only after the folder is there, since `ModLibrary.PrepareManifest` removes non-core entries whose folder is missing and the entry would be gone after the next launch. |
| The file is read once and written back. | It is re-read before every write, because `ModManifest.Save` regenerates it from the game's in-memory list and Borea has to expect it to change underneath. |
| Reading assumes Borea wrote the file. | Reading tolerates what the game and a human write: quoting, comments, and ids differing only in case. `ModIds.Comparer` is already the right comparer for the lookup. |

One thing that follows and is worth stating plainly: Borea can keep no state of its own in this file, because `ModManifest.Save` emits only `id` and `enabled` per entry and destroys everything else.

## Tests

One per rule, each naming the game behavior it protects against.

## One small thing to clean up while in there

`ManifestDto` and `ModManifestEntryDto` carry `System.Text.Json` attributes while `TomlFileStore` serializes with Tomlyn, which does not read them. They have no effect and suggest a contract that is not there.

Maximilian-Nesslauer:
Closes #53.

`FileModStateRepository` appended an entry for any id it was asked to enable, flipped entries it had not written, and read the file as if Borea had written it.

It follows the game's rules now: entry order is load order and only `ReorderAsync` changes it, `AddEntryAsync` is the only place an entry is created and writes nothing while the mod's `mod.toml` is missing, an entry that already exists is never flipped, and a lookup reaches what the game and a human write.

Removing an entry stays out. The game's startup sweep drops one whose files are gone, and reconciling folders against entries is #52.

### DIFF
--- a/src/Borea.Storage/State/FileModStateRepository.cs
+++ b/src/Borea.Storage/State/FileModStateRepository.cs
@@ -1,4 +1,4 @@
-﻿using Borea.Core.Mods;
+using Borea.Core.Mods;
 using Borea.Core.Paths;
 using Borea.Core.State;
 using Borea.Storage.Toml;
@@ -6,11 +6,15 @@ using Borea.Storage.Toml;
 namespace Borea.Storage.State;
 
 /// <summary>
-/// File-backed implementation of IModStateRepository, reading and writing
-/// the instance's manifest.toml directly.
+/// File-backed <see cref="IModStateRepository"/> over the instance's
+/// manifest.toml. ModManifest.Save regenerates that file from the game's own
+/// list and emits only id and enabled, so Borea keeps nothing of its own in it
+/// and re-reads before every write.
 /// </summary>
 public sealed class FileModStateRepository : IModStateRepository
 {
+    private const string ModDefinitionFileName = "mod.toml";
+
     private readonly IGamePathProvider _pathProvider;
 
     public FileModStateRepository(IGamePathProvider pathProvider)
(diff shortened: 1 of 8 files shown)


src/Borea.Storage/State/FileModStateRepository.cs
```diff
@@ -1,60 +1,210 @@
...
+            return ModEntryAddResult.AlreadyListed;
+
+        // the game builds paths from the id
+        // (ModEntry.Exists) and compares ids ordinally (ModLibrary.AddMods), so a
+        // case mismatch is a duplicate entry on Windows and a deleted one
+        // elsewhere. Valid by construction, since it equals a validated id apart
+        // from case and every ModIds rule ignores case.
+        var folderName = Path.GetFileName(modFolder);
+
+        manifest.Mods.Add(new ModManifestEntryDto { Id = folderName, Enabled = enabled });
```
MrJeranimo, line 74: Are we using the folder name because some OS file systems use case sensitive comparison and some don't. So this is just extra protection for this, even though the `modId` should be valid on all OSes.

src/Borea.Storage/State/FileModStateRepository.cs
```diff
@@ -1,60 +1,210 @@
...
+        {
+            if (enabledMatches.Count == 0)
+                return false;
+
+            // All of them, because any enabled entry loads the mod.
+            foreach (var match in enabledMatches)
+                match.Enabled = false;
+        }
+
+        await WriteManifestAsync(instanceId, manifest, cancellationToken).ConfigureAwait(false);
```
MrJeranimo: **[Major]**
I am not sure if I am correctly understanding this function, but it seems as if `Manifest.Mods` is never having the entry written back to, unless `matches` contains references and updates to matches update the original object inside `Manifest.Mods`.

If I am correct, this function never actually saves the changes to the manifest.

MrJeranimo (approved):

src/Borea.Storage/State/FileModStateRepository.cs
Maximilian-Nesslauer: Yes.
`ModIds` compares case-insensitively, per RFC 0031, so a caller can hand us `advancedflightcomputer` while the folder on disk is `AdvancedFlightComputer`, and both are correct input. I think that allowing that is better for "lazy" users that dont want to 100% type the correct spelling.

The game is stricter than we are:

- `ModLibrary.AddMods` decides whether a folder is already listed with `entry.Id == modDir.Name`, an ordinal comparison. A case mismatch is therefore never a match, and the game adds a second entry for the real folder name, new and disabled.
- `ModEntry.Exists` builds the path from the entry id instead, so there the file system decides.

That gives a different failure on each platform:

| Platform | What a case mismatch does |
|---|---|
| Windows | `Exists` is true, so our entry stays, and `AddMods` adds the folder entry next to it. Two entries for one mod, one enabled and one disabled, re-created on every launch. |
| Linux (and i believe also macos but not sure since i cant test it there but it seems like it *can* be case-senstive based on volume?) | `AddMods` adds the folder entry disabled, then `PrepareManifest` removes ours because `Exists` is false. The mod the user enabled is disabled after one launch. |

Also `ModLibrary.PrepareAll` passes the manifest entry's id into `Mod.MakeUsing`, so the id the mod carries at runtime is what the manifest says, not what the folder is called.

So the folder name is afaik the only spelling that survives a launch across all OSs.

Issue #65, opened 2026-09-01
Add a command line interface next to the desktop app


Maximilian-Nesslauer:
## current status

`Borea.App` is the only executable, and it is still the Avalonia template.

A command line interface was named in Discord on 2026-08-06 as something we want next to the desktop app, and no issue records it.

## Why

- Reproducible installs from a script, and later a modded server on a headless host once the game has multiplayer.
- A second consumer of `Borea.Core`, `Borea.Storage` and `Borea.Network` proves that no logic lives in the UI. The planned in-game mod needs that separation anyway.
- The libraries do the work, so the CLI is a thin wrapper and can exist before the app UI is done. It doubles as the test tool for the install path in #6.
- A fallback where no GUI is available.

## Shape

- A `Borea.Cli` project under `src/`, a console `Exe`, referencing the three libraries through #64 and never `Borea.App`.
- System.CommandLine 2.0, which is the .NET standard for this and is stable on NuGet.
- Every command maps onto a Core operation that exists or has an issue. The CLI adds no logic of its own.
- Human-readable output by default, `--json` on every read command for scripts.
- Exit codes: 0 done, 1 the operation failed, 2 bad usage.
- Tests in `tests/Borea.Cli.Tests`, each command run against the per-test temp directory the storage tests use, asserting the exit code and the `--json` output.

## Slices

One sub-issue and one pull request each, opened when its dependency is in progress.

- [ ] #66 Slice 1: settings, game version, index refresh, instances, enable and disable.
- [ ] #67 Slice 2: search and show.
- [ ] #68 Slice 3: install, remove, update.

Not now: `launch`, `publish`, `amend`. No Core operation exists behind them yet.

## Depends on

#64.

Maximilian-Nesslauer:
All slices are complete: #66, #67, #68 and #103. The launch command came with #112.

Issue #50, opened 2026-08-29
Build the Borea.App shell against the design in #8


Maximilian-Nesslauer:
`Borea.App` is still the Avalonia template: `MainViewModel` carries a `Greeting` and `MainWindow.axaml` shows a text box bound to it.

Meanwhile #8 has produced a navigation model, wireframes, high-fidelity screens, a state model and an actions reference.

## obstacle

`description` is CommonMark per RFC 0031, and Avalonia's Markdown control is not free.

## Not in scope

Any action from the reference in #8 that needs an installer or a resolver behind it. Those come with their own issues, and this shell should have the surfaces ready for them.

PR #70, opened 2026-09-02
Add a composition root for Borea.App and Borea.Cli

### LINKED ISSUE #64: One composition root for Borea.App and Borea.Cli
## current status

No program composes the three libraries.

`Borea.App` references `Borea.Core` and `Borea.Storage` and not `Borea.Network`, and `App.axaml.cs` constructs `new MainViewModel()` with no services behind it.
`GamePathProvider` is constructed nowhere outside the tests.
`LatestVersionPing`, `SpaceDockModRepository`, `CompositeModRepository`, `FileInstanceRepository`, `FileModStateRepository` and the rest each take their dependencies through the constructor, and nothing wires them together.

The README says Borea is "modifiable by changing out `Borea.Storage`, `Borea.Network`, and `Borea.App`".
A composition root is the one place that swap happens, and today it does not exist, so the claim has nothing to point at.

Both executables need it: #50 for the app shell, #65 for the command line interface.
Written twice it drifts twice.

## desired status

- One place that turns the saved settings into every service: the path provider, the `HttpClient`, the settings repository, the instance repository, the manifest repository, the mod sources, the downloader, the uninstaller, the version ping, and the index fetcher from #60.
- It sits above `Borea.Storage` and `Borea.Network`, because neither references the other, so it is a small new library referencing all three, or it lives in each executable. A library, so it is written once.
- Plain factory or `Microsoft.Extensions.DependencyInjection` is the decision to make here.
- The bootstrap has two steps, and the code has to say so: the settings file lives under `%LocalAppData%\Borea` and needs no game path to be found, but `GamePathProvider` takes the game path in its constructor. So a first path provider without a game path reads the settings, and the real one is built from them.
- `HttpClient` is created once with a `User-Agent`, and the same instance reaches every network class, which the `LatestVersionPing` cache already assumes.
- Nothing in `Borea.Core` changes.

## Tests

The root builds against the per-test temp directory the storage tests use, with no settings file, with a settings file naming no game, and with a full one.
No test reaches the network.

## Not in scope

Any UI, and any command. Those are #50 and #65.

## Depends on

#58 for the settings shape, #60 for the fetcher.

Maximilian-Nesslauer:
Closes #64.

Stacked on #58, because the settings shape comes from there.

One new library, `Borea.Composition`, turns the saved settings into every service, once. `BoreaServices.BuildAsync` bootstraps in two steps: a `GamePathProvider` without a game path reads the settings, and the real one is built from them. One `HttpClient` with a `User-Agent` reaches every network class.

`Borea.Core` is untouched.

The App builds the services in `Main` and hands them to `App`. No UI changes. `App.Services` is where the App shell picks them up.

Known, and left for follow-ups:
- A settings file that does not load ends the App before a window shows. Surfacing that is UI and belongs to #50.
- `SpaceDockResolver` maps ids per process, so a fresh process cannot resolve an installed mod's SpaceDock id. It needs persistence next to the installed-mod records, which matters for #65.
- Changed settings need a rebuilt graph. `Main` owns it, so a first run needs a restart after the game folder is picked, until #50 wires a rebuild.

### DIFF
--- a/Borea.slnx
+++ b/Borea.slnx
@@ -1,11 +1,13 @@
 <Solution>
   <Folder Name="/src/">
     <Project Path="src/Borea.App/Borea.App.csproj" />
+    <Project Path="src/Borea.Composition/Borea.Composition.csproj" />
     <Project Path="src/Borea.Core/Borea.Core.csproj" />
     <Project Path="src/Borea.Network/Borea.Network.csproj" />
     <Project Path="src/Borea.Storage/Borea.Storage.csproj" />
   </Folder>
   <Folder Name="/tests/">
+    <Project Path="tests/Borea.Composition.Tests/Borea.Composition.Tests.csproj" />
     <Project Path="tests/Borea.Core.Tests/Borea.Core.Tests.csproj" />
     <Project Path="tests/Borea.Network.Tests/Borea.Network.Tests.csproj" />
     <Project Path="tests/Borea.Storage.Tests/Borea.Storage.Tests.csproj" />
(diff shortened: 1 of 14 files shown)


MrJeranimo (approved):
LGTM

Issue #66, opened 2026-09-01
CLI slice 1: settings, game version, index refresh, instances


Maximilian-Nesslauer:
Part of #65, and the first pull request against #64.

## Commands

Everything here has a Core operation on `main` or in an open pull request.

| Command | Operation |
|---|---|
| `borea settings show` | `IBoreaSettingsRepository.GetAsync` |
| `borea settings set game <dir>` | `IBoreaSettingsRepository.SaveAsync` |
| `borea settings set loader <id> <dir>` | the loader map from #58 |
| `borea game version` | the installed build from #57, and the master server's build from `LatestVersionPing` |
| `borea index refresh` | `IContentIndexFetcher.FetchAsync` from #60, reporting downloaded or not modified |
| `borea instance list` | `IInstanceRepository.GetAllAsync`, marking the active one |
| `borea instance create <name>` | `CreateAsync` with `InstanceSource.Custom` |
| `borea instance rename`, `delete`, `activate` | `RenameAsync`, `DeleteAsync`, `SetActiveInstanceAsync` |
| `borea instance mods <instance>` | the manifest entries in load order, `GetEntriesAsync` from #62 |
| `borea enable <id> --instance <name>`, `borea disable` | `IModStateRepository.SetActiveAsync` and `SetInactiveAsync` |

`settings set` comes first, because without it nothing can point the CLI at a game.

An instance is named on the command line by its display name, and by id when two names differ only in case.

## Tests

One test per command, plus: no settings file, an instance name that is taken, an unknown instance, a disabled mod that stays disabled through `enable` of another mod.

## Not in scope

Anything from the index beyond fetching it, and anything that writes a mod folder.

## Depends on

#64, #57, #58, #60, #62.

PR #74, opened 2026-09-03
CLI slice 1: settings, game version, instances, enable and disable

Maximilian-Nesslauer:
Part of #66, stacked on #70.

**A second program next to the desktop app.** `src/Borea.Cli` builds an executable named `borea` on System.CommandLine 2.0.11.

It gets its services from `BoreaServices`, the composition root of #70, so the app and the command line always do the same thing.

Every command calls one operation that Borea.Core already offers.

**The commands in this slice.**
- `settings show`, `settings set game <dir>`, `settings set loader <id> <dir>`: where the game and the mod loaders are.
- `game version`: the current public build the master server reports.
- `instance list`, `create`, `rename`, `delete`, `activate`: the lifecycle of instances.
- `enable <mod-id>` and `disable <mod-id>`, with `--instance <name>` or the active instance when the option is absent.

**Two output modes.** Text for a person by default. `--json` on the three read commands for scripts.

**exit codes:** 0 when the command completed. 1 when the operation failed, with the reason on stderr. 2 when the command line did not parse, for example a missing argument or an invalid loader id.

**Not in this slice**, because their operations are not on the base branch yet: `index refresh` (#60) and `instance mods` (#62).

**To try it** after `dotnet build Borea.slnx`, with `borea` standing for `dotnet src/Borea.Cli/bin/Debug/net10.0/borea.dll`:
- `borea --help`
- `borea instance create Test`
- `borea enable SomeMod --instance Test`
- `borea instance list --json`
- `borea justSomeNonsense`, which exits 2

The executable uses `%LocalAppData%\Borea`, the same folder as the app.

LLM usage disclosure: Claude Fable 5.1 (and Opus 5 for an code review workflow) was used collaborativly with me to create large parts of this code. I read, reviewed and tested all of it myself and like it.

### DIFF
--- a/Borea.slnx
+++ b/Borea.slnx
@@ -1,12 +1,14 @@
 <Solution>
   <Folder Name="/src/">
     <Project Path="src/Borea.App/Borea.App.csproj" />
+    <Project Path="src/Borea.Cli/Borea.Cli.csproj" />
     <Project Path="src/Borea.Composition/Borea.Composition.csproj" />
     <Project Path="src/Borea.Core/Borea.Core.csproj" />
     <Project Path="src/Borea.Network/Borea.Network.csproj" />
     <Project Path="src/Borea.Storage/Borea.Storage.csproj" />
   </Folder>
   <Folder Name="/tests/">
+    <Project Path="tests/Borea.Cli.Tests/Borea.Cli.Tests.csproj" />
     <Project Path="tests/Borea.Composition.Tests/Borea.Composition.Tests.csproj" />
     <Project Path="tests/Borea.Core.Tests/Borea.Core.Tests.csproj" />
     <Project Path="tests/Borea.Network.Tests/Borea.Network.Tests.csproj" />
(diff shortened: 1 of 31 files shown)


PlazmaBoltz (approved):
All looks good, however is there a way to modify settings from internal code? #50 currently produces its own JSON within Borea.App to store system default themes and custom themes. When this is merged #50 will need a way to directly modify settings from the UI.

Maximilian-Nesslauer:
> All looks good, however is there a way to modify settings from internal code? #50 currently produces its own JSON within Borea.App to store system default themes and custom themes. When this is merged #50 will need a way to directly modify settings from the UI.

Thank you very much for reviewing the PR :)

#74 provides persistence for Borea’s game and loader paths.

Themes should use a separate app-preferences store, though since they do not affect the Core service graph, and BoreaSettings intentionally excludes app-level preferences. I did not want to add it here since the CLI slice 1 PR is already big enough for reviewers to read.

#50 currently only reads its JSON files, so persisting the selected theme and custom themes will need follow-up work in #50 or a separate issue. It should store them under Borea’s data directory, not in the packaged client folder.

#50 is IMO already big enough so i will create an new separate sub-Issue of #50 for it 👍

@PlazmaBoltz do you want to be assigned to the sub issue since you are already working on #50 ?

PR #75, opened 2026-09-03
Tidy the index fetcher

Maximilian-Nesslauer:
`snapshot_version` was read as a `uint` to keep negatives out and then cast back to `int`, and the cast wraps for anything above `int.MaxValue`, so the value came out negative and passed the ceiling:

```text
snapshot_version = 3000000000  ->  (int) = -1294967296  ->  accepted, cached copy replaced
snapshot_version = 4294967295  ->  (int) =          -1  ->  accepted, cached copy replaced
```

The rest:

- `SnapshotVersions.Highest` replaces the hardcoded `1` and its two "change this later" comments. Useful for #44
- The body is already in memory, so it is parsed directly instead of through an undisposed `MemoryStream`, which also lets the check stop being `async`.
- `out _` for the three properties that are only checked for presence.
- `<GenerateAssemblyInfo>false</GenerateAssemblyInfo>` comes back out of `Borea.Storage.Tests`. The duplicate attributes it was aimed at come from the fixture projects sitting inside the test project folder, and the `Compile Remove` on `Fixtures\**` from #57 already answers that. I removed the flag with both fixture `obj/**/*.AssemblyInfo.cs` files on disk.

### DIFF
--- a/src/Borea.Core/Index/SnapshotVersions.cs
+++ b/src/Borea.Core/Index/SnapshotVersions.cs
@@ -1,0 +1,15 @@
+namespace Borea.Core.Index;
+
+public static class SnapshotVersions
+{
+    /// <summary>
+    /// The highest snapshot version this build reads.
+    /// </summary>
+    public const int Highest = 1;
+
+    /// <summary>
+    /// Whether the snapshot is from a newer format than this build implements.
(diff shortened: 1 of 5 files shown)


MrJeranimo (approved):
Thanks for catching the things I missed

PR #77, opened 2026-09-04
Launch an instance through the mod loader

### LINKED ISSUE #69: Launch an instance through the mod loader
## current status

Nothing in Borea starts the game.

`Instance` says the instance is "redirected at launch via CLA/ENV path substitution", `IGamePathProvider` says the loader path is "the base path for launching StarMap", and no code does either.

## desired status

| Today | Should be |
|---|---|
| No launcher. | `ILauncher` in `Borea.Core`, implemented outside it, following the `ILatestVersionPing` precedent of interface in Core and process handling outside it. |
| The instance folder never reaches the loader. | The launch passes `-InstancePath <absolute instance root>`. StarMap's `DocumentsPathPatches` reads it before mod discovery, so the manifest, mods, saves, settings and logs of that instance are what the game sees (research/starmap.md). `STARMAP_INSTANCE_PATH` is the equivalent environment variable. |
| Nothing knows what to run. | `[provides].launch` of the loader's authored listing names the executable (#59 carries it), resolved against the loader directory from #58. It is read from the live listing every time and never from a cached release file, because RFC 0035 never stamps `[provides]`. |
| No working directory rule. | The process starts with its working directory set to the loader's install location, RFC 0035 rule 5, because `LoaderConfig.TryLoadConfig` reads `StarMapConfig.json` relative to the process working directory. |
| Borea cannot tell whether the game runs. | The running process is `StarMap.exe`, not the game (research/starmap.md). Borea refuses to write into an instance's mods folder or manifest while a launch it started is still running, and says so. Detecting a StarMap the user started by hand is a follow-up. |
| A loaderless launch is silently the shared profile. | Launching `KSA.exe` directly stays out: the game takes no path argument and reads no environment variable, so no instance can be selected without a loader (research/ksa-mod-loading.md). Borea reports that instead of launching against the shared profile. |

## Tests

A fake process starter that records the executable, the arguments, the working directory and the environment.

Cases: an instance with a known loader, no loader path configured, a listing whose `[provides]` has no `launch`, and a second launch while the first one is still running.

## Not in scope

Reading the game log after it exits (content-manager-design#43), instances per game build (content-manager-design#30 rules it out), and driving StarMap's own restart flow.

## Depends on

#58 for the loader path, #59 for `[provides].launch`.

Maximilian-Nesslauer:
Closes #69.

**Borea can start the game for an instance now.** A launch runs like this:

1. The caller says: start instance X with the loader StarMap.
2. Borea looks up in the settings where StarMap is installed.
3. Borea reads from the live StarMap listing which file to run. That is `[provides].launch`, so `StarMap.exe`.
4. Borea starts that file inside the StarMap folder, so StarMap finds its `StarMapConfig.json`, and hands over the instance folder the way the loader takes it. For StarMap that is the argument `-InstancePath <path>` and the environment variable `STARMAP_INSTANCE_PATH`.
5. StarMap points the game's Documents path at that folder. Mods, `manifest.toml`, saves, settings and logs all come from the instance.
6. Borea remembers that the instance is running. A second launch of the same instance is refused until the process ends.

**When something is missing** there is one sentence for the user: no loader chosen, no folder in the settings, a listing without `launch`, the executable is not there, or the system refused to start it. Without a loader nothing is started, because the game alone reads no instance path and would land in the shared profile.

**The one thing the format does not say yet** is how a loader takes an instance. RFC 0035 leaves that to discussion #20 and `[provides]` has no field for it. So Borea keeps a small table of what it knows, and that is StarMap only: the flag, plus the variable, because StarMap's own restart passes only `--restarted` and keeps the environment. Any other loader is refused with a message instead of started with a guess. Once the format gains a field for it, the table goes away and the listing says it.

**Tested against the real game.** StarMap printed the instance path, and settings, manifest, mods, saves and logs all landed in the instance folder. The Documents profile stayed untouched.

**Not in this pull request:** the one line in `BoreaServices` (#70), the installer and the manifest writer asking `IsRunning` before they write (after #72), a CLI `launch` command, detecting a StarMap the user started by hand, and the format field for the handover, which is a design repo question.

LLM usage disclosure: Claude Fable 5.1 (and Opus 5 for an code review workflow) was used collaborativly with me to create parts of this code. I read, reviewed and tested all of it myself and like it.

### DIFF
--- a/src/Borea.Core/Launch/ILauncher.cs
+++ b/src/Borea.Core/Launch/ILauncher.cs
@@ -1,0 +1,19 @@
+using Borea.Core.Instances;
+using Borea.Core.Mods;
+
+namespace Borea.Core.Launch;
+
+/// <summary>
+/// Starts the game for one instance through its mod loader.
+/// </summary>
+public interface ILauncher
+{
+    LaunchResult Launch(Instance instance, ModMetadata? loader);
+
+    /// <summary>
+    /// Whether a launch this launcher started for the instance is still
+    /// running. A launch from an earlier session, a game started by hand, and
+    /// the new process a loader starts when it restarts itself are not seen.
+    /// </summary>
+    bool IsRunning(Guid instanceId);
+}
--- a/src/Borea.Core/Launch/InstanceHandover.cs
+++ b/src/Borea.Core/Launch/InstanceHandover.cs
@@ -1,0 +1,47 @@
+using Borea.Core.Mods;
+
+namespace Borea.Core.Launch;
+
+/// <summary>
+/// How a loader is told which instance to run: a flag that takes the instance
+/// root as the next argument, a variable that carries it, or both.
+/// </summary>
+public sealed class InstanceHandover
+{
+    // StarMap reads the flag first and the variable when the flag is absent.
+    // Its own restart passes only --restarted, so the variable is what keeps
+    // the instance across it.
+    private static readonly Dictionary<string, InstanceHandover> KnownHandovers = new(ModIds.Comparer)
+    {
+        ["StarMap"] = new InstanceHandover("-InstancePath", "STARMAP_INSTANCE_PATH"),
+    };
+
+    /// <summary>The argument that precedes the instance root. Null when the loader takes no flag.</summary>
+    public string? Flag { get; }
+
+    /// <summary>The environment variable that carries the instance root. Null when the loader reads none.</summary>
+    public string? Variable { get; }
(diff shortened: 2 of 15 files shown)


Maximilian-Nesslauer:
i just posted

https://github.com/KSAModding/content-manager-design/discussions/20#discussioncomment-18285038

which resulted from this PR

PlazmaBoltz (approved):

PR #85, opened 2026-09-04
Install a mod loader per RFC 0035

### LINKED ISSUE #51: Install a mod loader per RFC 0035
A code mod cannot run without an mod loader.

## Where this stands today

`BoreaSettings` carries a single `StarMapDirectoryPath`, mirrored by `IGamePathProvider.GetStarMapDirectoryPath`.
The getter has no caller outside the tests, so nothing installs a loader and nothing configures one.

## What it should be instead

| Today | Should be |
|---|---|
| One hardcoded StarMap path. | A path per loader id, which #42 introduces, and a real install that puts the files there. |
| No notion of where a loader goes. | The `[install]` anchor decides: `standalone` is a directory Borea chooses, `mods` is the game's mods folder, `user-data` is the user data root, `game-root` is the folder holding the game executable and `Content`. |
| The user points Borea at a StarMap they installed themselves. | Borea unpacks the release into the resolved anchor and records the location. |
| Nothing writes the loader's configuration, so the user edits `GameLocation` by hand. | `[provides.configure]` is written: only the keys the table names, everything else in the file preserved, and a missing file created with just those keys, which is enough for a loader that writes its own template on first run. |
| Nothing knows what to launch. | `[provides].launch` names it, and it is read from the authored listing every time rather than from a cached release file, because RFC 0035 deliberately never stamps `[provides]` and a stale copy would point at a directory the installed loader no longer reads. |

## Tests

Against the real StarMap listing, including a `StarMapConfig.json` that already holds other keys and has to survive the write.

## Depends on

`#42` for the per-loader paths and `#43` for the authored tables.

Maximilian-Nesslauer:
Closes #51. Stacked on #72, so the base is `install-mod`: the archive download and the unpack with the stamped root come from there.

**Borea installs a mod loader itself now, from its listing in the index, and points it at the game.** Until now a user downloaded StarMap by hand, unpacked it somewhere, edited its config file, and typed the folder into Borea's settings. Borea did not know anything about StarMap in particular before, and it still does not: it reads what the listing says and follows it. A second loader with a listing of its own works the same way without new code.

What happens when Borea installs a loader, in six steps:

1. **Check the two documents.** The listing (the author's file) and the release (the stamped file with the download link) must both describe a mod loader with the same id.
2. **Work out where it goes.** The StarMap release says "standalone", so Borea picks a folder of its own, `Loaders/StarMap` in its data folder. The caller can name another folder. A loader that says "below the game folder" goes there instead.
3. **Download.** The zip is fetched and its hash checked, with the same code #72 uses for mods.
4. **Unpack and check.** The zip is unpacked into that folder. Then Borea checks that the file the listing says to run, `StarMap.exe`, is really there (RFC 0035 rule 3). If not, the install fails and the folder is removed again.
5. **Configure.** Borea writes the game path into the loader's own config file. It writes only the one key the listing names, `GameLocation`, and leaves every other line alone. A file that does not exist yet is created with just that key.
6. **Record.** Borea writes into its settings that StarMap is installed in this folder. The launcher (#77) reads that to start the game.

**Installing again.** A loader that is already recorded is installed into the same folder, over the old files. The config file is kept as it is, even when the new zip ships a fresh one, because the user may have edited it. Then the game path key is written again.

**What Borea refuses.** A folder that already holds files Borea did not put there. A loader that wants to go directly into the game folder, because those files could never be taken out again. The two anchors that mean "inside an instance", because Borea remembers one folder per loader and not one per instance. And an install while the game directory is not set, because then the config cannot be filled in.

**Not in this pull request:** adopting a StarMap the user installed by hand, uninstalling a loader, the wiring into the composition root (#70) and the launcher (#77), and rewriting the game path in every loader when the game directory changes. The path provider built at startup does not see the new record until a restart, which the composition root owns.

LLM usage disclosure: Claude Fable 5.1 (and Opus 5 for the code review workflow) was used collaboratively with me to create parts of this. I read, reviewed and tested all of it myself and like it.

### DIFF
--- a/src/Borea.Core/ModLoaders/ILoaderConfigurator.cs
+++ b/src/Borea.Core/ModLoaders/ILoaderConfigurator.cs
@@ -1,0 +1,27 @@
+using Borea.Core.Mods;
+
+namespace Borea.Core.ModLoaders;
+
+/// <summary>
+/// Writes the loader's own configuration file per <c>[provides.configure]</c>
+/// of RFC 0035.
+/// </summary>
+public interface ILoaderConfigurator
+{
+    /// <summary>
+    /// Writes the absolute game directory to the key <c>game-path</c> names, in
+    /// the file below <paramref name="loaderDirectory"/>, and returns the
+    /// file's absolute path, or null when the listing names no key.
+    /// </summary>
+    /// <exception cref="ArgumentException">The listing is not a mod loader, or a directory is not absolute.</exception>
+    /// <exception cref="InvalidOperationException">
+    /// The file cannot be read as its stated format, holds a key twice, or a
+    /// value sits where the key path expects a table.
+    /// </exception>
+    /// <exception cref="NotSupportedException">The listing names a format this configurator cannot write.</exception>
+    Task<string?> ConfigureAsync(
+        ModMetadata loader,
+        string loaderDirectory,
+        string gameDirectory,
+        CancellationToken cancellationToken = default);
+}
--- a/src/Borea.Core/ModLoaders/ILoaderInstaller.cs
+++ b/src/Borea.Core/ModLoaders/ILoaderInstaller.cs
@@ -1,0 +1,55 @@
+using Borea.Core.Mods;
+
+namespace Borea.Core.ModLoaders;
+
+/// <summary>
+/// Installs one release of a mod loader per RFC 0035 and records its
+/// directory in the settings. Starting it is the launcher's job.
+/// </summary>
+public interface ILoaderInstaller
+{
+    /// <summary>
+    /// The stamped install decides where the loader goes; the live listing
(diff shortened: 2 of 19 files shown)


averageksp (commented):
I took a look and the PR #85 looks solid overall, but I’m not ready to approve it yet because replacement installs can still leave partial state on failure. In FileLoaderInstaller.InstallAsync, the catch block intentionally doesn’t fully roll back when replacing == true, so if unpacking, config restore/write, or saving settings fails, the loader folder can end up half-updated without the settings matching it.

If that behavior is intentional, I’d love to see it called out clearly in the PR. Otherwise, I think it would be safer to make replacement installs transactional too.
<@315123705045057536>

averageksp (approved):

PR #89, opened 2026-09-08
Added JSON Data Transfer Objects to use a JSON serializer for easy validation and data transfer.

MrJeranimo:
This Is just the DTOs. I am bsically looking for advice if these DTOs match the Index shape and if the failure pattern is what we want.

From Root down: 
- `SnapshotDto` -> if SnapshotVersion, GameVersions, Listings, or Packs are missing or the wrong object, the whole validation will fail. If Sources is missing it is ok.
- `GameVersions` -> Must be fully parseable or whole validation fails (should be ok since it is machine generated)
- `Listings` and `Packs` -> If one listing/pack fails, only that listing/pack fails. Continues validation for the rest of the listings/packs.
- ` ListingEntryDto` -> If Id missing or fails, whole listing gone. Checks for, if Id and no authored/releases, then needs IndexStatus will come later. If one release for the listing fails, only that release fails, not the whole listing.
- `AuthoredDto` -> Requires all fields listed in the CMD, `CompatabilityDto` is required since `game_min` is required. `LinksDto` is required since `forums` link is required. `Releases` fails silently as a broke `[releases]` object shouldn't break the whole authored listing. `Loader` fails loudly as listing that declares a loader but can't parse it means it can't reasonable by determined/trusted. `Dependencies` fails loudly for the same reason. Hence `Loader` and `Dependencies` failing will cause `AuthoredDto` to fail and thus the listing to fail.
- `ReleasesEntryDto` -> Follows CMD to the best I could find. Most properties are nullable since a yanked release will make them null. `Dependencies`, `Loader`, `InstallInfo`, and `ReleaseListing` all fail loudly as broken data on those will mean a broken release.
- `PackEntryDto` -> uses `AuthoredDto` for the Versions as they are almost exactly the same with the Packs version just not having nullable fields so it should work.
- `LinksDto` -> Requires a Forums link in the object. Uses a `Dictionary<string, JsonElement>` with a `[JsonExtensionData]` tag to  grab the rest of the links and allow custom links. The actual links will be extacted in the mapper.

Please let me know about any glaring errors. Enums are stored as string till DTO to actual object transfer. The Mappers will transfer the `List<JsonElement>` into `List<type>` and will be added later. The `JsonElement` substitution is to prevent one wrong listing from breaking everything. I added the `[JsonPropertyName]` tags to everything for extra clarity.

Please let me know what you think about these, I have tried many things and this is what I thought would be best, so any extra info, viewpoints, etc. will be greatly appreciated.


src/Borea.Storage/Index/IndexValidator.cs
MrJeranimo, line 1: Ignore this file. This was how I was doing the validation before. It hasn't been updated to the DTO specs. It was also incomplete so it is causing the build to fail. I didn't mean to add it.

src/Borea.Storage/Index/LoaderDto.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System.Text.Json.Serialization;
+
+namespace Borea.Storage.Index;
+
+public sealed class LoaderDto
+{
+    [JsonPropertyName("id")]
+    public required int Id { get; set; }
```
Maximilian-Nesslauer: [must fix] `loader.id` is a content id string, not a number. The live index writes `id = "StarMap"`, so this type rejects AdvancedFlightComputer, KSArmory, and MeasureTools before their other fields can be read. Please change this to `string` and let `LoaderRequirement` validate the id.

src/Borea.Storage/Index/ReleasesEntryDto.cs
```diff
@@ -0,0 +1,63 @@
...
+    [JsonPropertyName("download")]
+    public DownloadInfoDto? Download { get; set; }
+
+    [JsonPropertyName("install_size")]
+    public int? InstallSize { get; set; }
+
+    [JsonPropertyName("install")]
+    public InstallInfoDto? Install { get; set; }
+
+    [JsonPropertyName("loeader")]
```
Maximilian-Nesslauer: [must fix] The JSON key is misspelled as `loeader`. `System.Text.Json` therefore ignores every real `loader` block without an error.

src/Borea.Storage/Index/InstallInfoDto.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System.Text.Json.Serialization;
+
+namespace Borea.Storage.Index;
+
+public sealed class InstallInfoDto
+{
+    [JsonPropertyName("root")]
+    public required string Root { get; set; }
```
Maximilian-Nesslauer: [must fix] `root` is optional in a stamped install object.
The StarMap 0.4.6 release correctly contains `{ "derived": true, "target": "standalone" }` with no `root`, but this DTO rejects it. Please make `Root` nullable and add `Target` and `Path`, which are already part of `InstallInfo` and RFC 0035. Without `Target`, Borea does not know where to install a mod loader.

src/Borea.Storage/Index/AuthoredDto.cs
```diff
@@ -0,0 +1,59 @@
...
+    [JsonPropertyName("releases")]
+    public JsonElement? Releases { get; set; }
+
+    // Fails loudly if a loader is declared but can't be parsed
+    [JsonPropertyName("loader")]
+    public LoaderDto? Loader { get; set; }
+
+    // Fails loudly if dependencies are declared but can't be parsed
+    [JsonPropertyName("dependencies")]
+    public List<DependencyEntryDto>? Dependencies { get; set; }
```
Maximilian-Nesslauer, line 57: [must fix] This DTO does not include the accepted RFC 0035 `install` and `provides` tables.

The current StarMap listing uses both for its standalone target, launch executable, content directory, and configuration file. The listing can still be turned into a `ModMetadata`, but those important fields are silently lost. Please add DTOs for the complete current shapes. `provides.instance` can follow after RFC 0049 is accepted.

src/Borea.Storage/Index/ReleasesEntryDto.cs
```diff
@@ -0,0 +1,63 @@
...
+    [JsonPropertyName("type")]
+    public required string Type { get; set; }
+
+    [JsonPropertyName("version")]
+    public required string Version { get; set; }
+
+    [JsonPropertyName("version_scheme")]
+    public required string VersionScheme { get; set; }
+
+    [JsonPropertyName("release_date")]
```
Maximilian-Nesslauer, line 26: [must fix] A yanked release is still a complete release. Yanking only adds `yanked = true` and an optional reason, so it does not make the other fields nullable. `release_date`, `game_min_revision`, `download`, `install_size`, and `dependencies` are written on every stamped release. This DTO is also missing `release_status` and the optional `os` list. Please mirror the stamped shape and keep each release inside its own error boundary.

src/Borea.Storage/Index/ListingEntryDto.cs
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Borea.Storage.Index;
+
+public sealed class ListingEntryDto
+{
+    [JsonPropertyName("id")]
+    public required string Id { get; set; }
+
+    [JsonPropertyName("authored")]
+    public AuthoredDto? Authored { get; set; }
```
Maximilian-Nesslauer, line 12: [must fix] Please read the outer `id` before you deserialize the `authored` block. If `ListingEntryDto` is deserialized in one step, any error inside `AuthoredDto` prevents the caller from getting `Id` from the result. The reader must still return the listing id and an error reason because an unreadable entry keeps its id reserved. Keeping `authored` as a `JsonElement?` until the second step would make this boundary clear.

src/Borea.Storage/Index/DownloadInfoDto.cs
```diff
@@ -0,0 +1,18 @@
...
+public sealed class DownloadInfoDto
+{
+    [JsonPropertyName("url")]
+    public required string URL { get; set; }
+
+    [JsonPropertyName("sha256")]
+    public required string SHA256 { get; set; }
+
+    [JsonPropertyName("size")]
+    public required int Size { get; set; }
```
Maximilian-Nesslauer: Please use `long` for `size` and `install_size`, as the Core model already does. 

This DTO also needs the optional `mirrors` array, which is already represented by `DownloadInfo.Mirrors`.

src/Borea.Storage/Index/ReleasesInfoDto.cs
```diff
@@ -0,0 +1,15 @@
...
+
+public sealed class ReleasesInfoDto
+{
+    // If exists, must have an authority
+    [JsonPropertyName("authority")]
+    public required string Authority { get; set; }
+
+    // if exists, must have at least one other element besides authority
+    [JsonExtensionData]
+    public required Dictionary<string, JsonElement> Hosts { get; set; }
```
Maximilian-Nesslauer: [must fix] `required` cannot be used on a `[JsonExtensionData]` property. `System.Text.Json` throws an `InvalidOperationException` for this type before it reads any input. `authority` is also optional when a listing names only one release host. Because this DTO is not used yet, I would remove it for now or fix both rules before `AuthoredDto.Releases` uses it.

src/Borea.Storage/Index/SnapshotDto.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// Same reason as Listings
+    /// </summary>
+    [JsonPropertyName("packs")]
+    public required List<JsonElement> Packs { get; set; }
+
+    /// <summary>
+    /// Chose to let the serializer throw if GameVersions is incorrect
+    /// </summary>
+    [JsonPropertyName("game_versions")]
+    public required GameVersionsDto GameVersions { get; set; }
```
Maximilian-Nesslauer, line 32: [must fix] `required` checks that these keys exist, but the default serializer options still accept explicit `null` values for the reference-type members.

The DTO also does not apply the `SnapshotVersions.Highest` ceiling by itself.

Please put the complete envelope check in one shared operation that verifies the root object, snapshot version, listings array, packs array, and game versions object. `ContentIndexFetcher` must run it on the downloaded bytes before it replaces the known-good cache, and Storage must run it again when it reads the cache at startup.

src/Borea.Storage/Index/PackEntryDto.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Borea.Storage.Index;
+
+public sealed class PackEntryDto
+{
+    [JsonPropertyName("id")]
+    public required string Id { get; set; }
+
+    [JsonPropertyName("versions")]
+    public List<JsonElement>? Versions { get; set; }
```
Maximilian-Nesslauer, line 13: [should fix] A pack version is wrapped as `{ "authored": ... }`, with an optional `index_status` for that version. The authored pack document also has a different shape from `AuthoredDto`: it requires `version`, `released_at`, and `mods`, and it can contain `vehicles`, `saves`, and `changelog`. Please add a DTO for the version wrapper and a separate DTO for the pack document. The live index has no packs yet, but the current shape would lose the fields that make the document a pack.

src/Borea.Storage/Index/ReleasesEntryDto.cs
```diff
@@ -0,0 +1,63 @@
...
+    public LoaderDto? Loader { get; set; }
+
+    [JsonPropertyName("dependencies")]
+    public List<DependencyEntryDto>? Dependencies { get; set; }
+
+    [JsonPropertyName("changelog")]
+    public string? Changelog { get; set; }
+
+    [JsonPropertyName("listing")]
+    public required ReleaseListingDto Listing { get; set; }
```
Maximilian-Nesslauer: [should fix] RFC 0031 lets a client fall back to the live authored data when the frozen `listing` block or one of its fields is absent. `ModVersionMetadata.Listing` is nullable for the same reason. Please keep this display-only block optional and decide how a partial block falls back before it is mapped.

src/Borea.Storage/Index/AuthoredDto.cs
```diff
@@ -0,0 +1,59 @@
...
+
+    [JsonPropertyName("compatibility")]
+    public required CompatibilityDto Compatibility { get; set; }
+
+    [JsonPropertyName("links")]
+    public required LinksDto Links { get; set; }
+
+    // Fails silently as a broken [releases] shouldn't break the whole Authored section
+    [JsonPropertyName("releases")]
+    public JsonElement? Releases { get; set; }
```
Maximilian-Nesslauer: A missing `releases` table and a broken one have different meanings.

Missing means that releases enter the index through pull requests.

If a present table cannot be read, the listing should become unreadable instead of silently changing to that fallback path. Please parse it inside the per-listing error boundary and keep the outer id and the reason when it fails.

Maximilian-Nesslauer (commented):
Thank you very much for the work :) It generally looks good and i left some in-line comments for what i would change.

Thanks for keeping the index DTOs separate from the immutable Core model.
I believe that Using `JsonElement` at the listing, release, and pack-version boundaries is the right approach because one broken entry must not remove the rest of the catalogue.
Keeping enum-like values as strings for `MetadataEnumMapper` also makes sense.

I teste these DTOs with a snapshot built from the current index repositories.
Three of the four listings fail because `loader.id` has the wrong type, and the StarMap release fails because its valid install object has no `root`.
Several fields would also be lost without an error, including authored `install` and `provides`, release `loader`, `release_status`, `install.target`, and download mirrors.

The error boundaries need one more step.
Read the outer id first, then deserialize and map the entry inside its own `try` block.
On failure, return that id and the error reason next to the usable entries.
A newer `spec_version` belongs in that unreadable list, while an unknown `index_status.state`, a tombstone, and an empty release list are valid cases.

Please also remove `IndexValidator.cs`, as you already noted.
The current branch does not compile because that file references an `IndexStatus` type that does not exist.
I believe that the snapshot-version ceiling is  useful, but the full envelope check must run before the fetcher replaces the old cache and again when Storage reads the cache after a restart.

Before the pull request leaves draft, I would add one fixture test that reads a snapshot built from the current index documents.
That single test catches most of the shape problems above.

I would then add focused cases for a tombstone, an empty release list, one bad listing beside a good one, one bad release beside a good one, an unknown index status, a newer spec version, a complete yanked release, and a retracted pack version.

src/Borea.Storage/Index/AuthoredDto.cs
MrJeranimo: My thought process on making `releases` fail silently is that Borea was unlikely to use this part of the listing, so I thought it would be better to not fail the listing. However, I am now thinking that maybe you are right and it should fail loudly since we may add something to Borea that uses the `releases` table and this part failing silently will be a hard bug to catch.

src/Borea.Storage/Index/ListingEntryDto.cs
MrJeranimo: I will prbably do that in the mapper, essentially check for the Id since in `SnapshotDto` Listings is a list of `JsonElement` so I have to make a custom/helper mapper for the `ListingEntryDto` where I can first grab the Id then try to serialize the document so if it fails I know why and for what Id.

src/Borea.Storage/Index/AuthoredDto.cs
Maximilian-Nesslauer: Yes, that is the result I meant.

It should fail inside the per-listing boundary, so only that listing becomes unreadable and the rest of the snapshot stays usable.

src/Borea.Storage/Index/InstallInfoDto.cs
MrJeranimo: I also am adding the optional `manages`, `steps`, and `uninstall` parts of the `[install]` table as RFC 0035 declares them.

Maximilian-Nesslauer: Yes, that makes sense.

Please keep them nullable so the DTO preserves the difference between a missing field and an empty array.
The path and content rules can be checked later when the DTO is mapped to the model.

src/Borea.Storage/Index/SnapshotDto.cs
MrJeranimo: > `required` checks that these keys exist, but the default serializer options still accept explicit null values for the reference-type members.

There is a `JsonSerializerOptions` option that lets you follow the compiled `?`/not-`?` rules so I will use that while deserializing the index.

```csharp
JsonSerializerOptions Options = new()
{
    RespectNullableAnnotations = true
};
```

> The DTO also does not apply the SnapshotVersions.Highest ceiling by itself.

I will check that before full deserialization. Same with all the other spec versions contained within the index.

> Please put the complete envelope check in one shared operation that verifies the root object, snapshot version, listings array, packs array, and game versions object. ContentIndexFetcher must run it on the downloaded bytes before it replaces the known-good cache, and Storage must run it again when it reads the cache at startup.

I plan to have a function separate from the full index validation `ValidateIndexRoot()` which only check the root level to check if everything required is there and if the types are correct. The full index validator will call that function, same with `ContentIndexFetcher`.

Maximilian-Nesslauer: That approach sounds good 👍 

`required` together with `RespectNullableAnnotations` should cover both missing properties and explicit `null` values.

One important boundary is that only `snapshot_version` should be checked before reading the complete snapshot. Please check the `spec_version` of each listing, release, and pack inside that entry's own error boundary, so an unsupported entry becomes unreadable without rejecting the rest of the index.

Please also make sure that `ContentIndexFetcher` calls `ValidateIndexRoot()` before it replaces the cached file, and that the cache reader calls the same function at startup.

MrJeranimo: I just realized that `ContentIndexFetcher` is in `Borea.Networking` while the `ValidateIndexRoot()` would be in `Borea.Storage`. For now, I will just make sure `ContentIndexFetcher` copies exactly what the `ValidateIndexRoot()` would do.

MrJeranimo:
I have updated the DTOs based on the feedback I received and checked the RFCs to see if I missed anything. I will now start on the actual validation and transfer code/logic.

One thing I want to mention, If there is a `spec_version`, `snapshot_version`, or `id` field for a type, I will manually extract and check that info before using the JSON serializer on the object type. This will allow for,

1. The serializer to only serialize objects and snapshots that are within the highest version.
2. Knowing which entry or part of the index failed to parse and why because its `id` is known.

Also anything with a `List<JsonElement>` as the type indicates that if one element of the list fails, ONLY that element fails, not the whole list.

src/Borea.Storage/Index/AuthoredDto.cs
```diff
@@ -0,0 +1,65 @@
...
+    // Fails loudly if a loader is declared but can't be parsed
+    [JsonPropertyName("loader")]
+    public LoaderDto? Loader { get; set; }
+
+    // Fails loudly if dependencies are declared but can't be parsed
+    [JsonPropertyName("dependencies")]
+    public List<DependencyEntryDto>? Dependencies { get; set; }
+
+    [JsonPropertyName("install")]
+    public InstallInfoDto? Install { get; set; }
```
Maximilian-Nesslauer: Use separate DTOs for authored and release install data

This DTO reads authored metadata, but `InstallInfoDto` requires `derived`, which only exists in a stamped release.

An authored install instead has `target`, `path`, `manages`, `steps`, and `uninstall`.

The current DTO rejects the live StarMap listing because its install table correctly has no `derived` field. Please use separate DTOs for the two wire formats. This also matches the existing split between `InstallDescriptor` and `InstallInfo` in Borea.Core.

src/Borea.Storage/Index/PackEntryDto.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace Borea.Storage.Index;
+
+public sealed class PackEntryDto
+{
+    [JsonPropertyName("id")]
+    public required string Id { get; set; }
+
+    // Uses PackVersionDto
+    [JsonPropertyName("versions")]
+    public List<JsonElement>? Versions { get; set; }
```
Maximilian-Nesslauer, line 13: Preserve the wrapper around each pack version

Each item in `versions` contains an `authored` object and an optional `index_status` object.

`PackVersionDto` currently expects the authored fields at the top level, so it cannot read a valid pack version from the snapshot.

Please add a small wrapper DTO and put `PackVersionDto` under its `authored` property. The `index_status` property belongs on that wrapper, not inside the authored document.

src/Borea.Storage/Index/ConfigureDto.cs
```diff
@@ -0,0 +1,15 @@
...
+public sealed class ConfigureDto
+{
+    [JsonPropertyName("file")]
+    public required string File { get; set; }
+
+    [JsonPropertyName("format")]
+    public required string Format { get; set; }
+
+    [JsonPropertyName("game-path")]
+    public string? GamePath { get; set; }
```
Maximilian-Nesslauer, line 15: Keep unknown configure members so the mapper can reject them

RFC 0035 makes `provides.configure` the exception to the normal unknown-field rule. An unknown member makes this table invalid because ignoring it could start a loader without a setting it needs. `System.Text.Json` currently discards unknown members here, so the mapper cannot enforce that rule. Please capture them with `JsonExtensionData`, or validate the raw object before this DTO is created.

src/Borea.Storage/Index/ReleaseListingDto.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System.Text.Json.Serialization;
+
+namespace Borea.Storage.Index;
+
+public sealed class ReleaseListingDto
+{
+    [JsonPropertyName("name")]
+    public required string Name { get; set; }
```
Maximilian-Nesslauer: Allow a release to freeze only some listing fields

The `listing` object in a stamped release can contain only the fields that changed. Missing fields fall back to the live authored metadata one field at a time. Requiring every core field here makes a valid partial object fail deserialization.

Please make these properties optional and let the mapper apply each present value over the authored listing. A present value with the wrong type must still make that release unreadable.

src/Borea.Storage/Index/SnapshotDto.cs
Maximilian-Nesslauer: Yes that seems fine for now but on the longterm I do not think we should copy the root validation into `ContentIndexFetcher`.

That would give us two implementations of the same rule, and they can drift when the snapshot format changes if we are not careful.

I suggest putting an interface such as `IContentIndexRootValidator` in `Borea.Core.Index`, implementing it once in `Borea.Storage.Index`, and injecting that implementation into both the fetcher and the index reader through `BoreaServices`. The fetcher can then validate the downloaded bytes before it replaces the cache, while the reader uses the same implementation when it opens the cached file at startup.

The shared check must reject invalid JSON, a non-object root, an unsupported `snapshot_version`, non-array `listings` or `packs`, and a non-object `game_versions`.

If you want to not do it in this PR it should be fine but please link this comment in your PR body so that we dont forget.

Maximilian-Nesslauer:
> I have updated the DTOs based on the feedback I received and checked the RFCs to see if I missed anything. I will now start on the actual validation and transfer code/logic.
> 
> One thing I want to mention, If there is a `spec_version`, `snapshot_version`, or `id` field for a type, I will manually extract and check that info before using the JSON serializer on the object type. This will allow for,
> 
>     1. The serializer to only serialize objects and snapshots that are within the highest version.
> 
>     2. Knowing which entry or part of the index failed to parse and why because its `id` is known.
> 
> 
> Also anything with a `List<JsonElement>` as the type indicates that if one element of the list fails, ONLY that element fails, not the whole list.

Extracting `snapshot_version`, `spec_version`, and the outer ids before deserialization makes sense in my opinion.

Only an invalid or unsupported `snapshot_version` rejects the complete snapshot. A bad authored document, release, or pack version stays local to that entry, and Borea should keep the outer id and the reason. For a bad release or pack version, it should also keep the raw version when that value can be read safely.

`List<JsonElement>` gives the reader the correct boundary, but it does not isolate failures by itself. The deserialize call and the model constructor for each element must both be inside that element's own error handling.

Also, `RespectNullableAnnotations` checks nullable properties, but it does not reject null values inside a `List<string>`. The mapper or model still has to check collection items and the semantic rules.

Maximilian-Nesslauer (commented):
The update fixes most of the DTO-shape problems from the first review, and the Storage project now compiles.

Before you start the mapper and validation logic, I think four points in the DTO layer still need changes:

- Authored install data and stamped release install data need separate DTOs because they have different fields.
- A pack version needs its `authored` and `index_status` wrapper.
- Unknown members in `provides.configure` must remain visible so Borea can reject that table as required by RFC 0035.
- A frozen release listing must allow any subset of listing fields and fall back to the authored listing field by field.

I would also keep the root validation in one shared implementation. Both the network fetcher and the cached index reader need the same rules, so copying the check into both projects will make future format changes harder to keep safe and maintain. See my comment on your comment.

I tested the DTOs against a snapshot built from the current local index data. Three listings and all nine of their releases deserialize, but StarMap fails because the authored install has no `derived` field. A valid pack-version wrapper also fails with the current `PackVersionDto`. I also confirmed that a null author and an unknown configure member pass through `System.Text.Json`, so those checks still need to happen outside its nullable-property handling.

src/Borea.Storage/Index/PackEntryDto.cs
MrJeranimo: I added the wrapper by changing `PackVersionDto` to be the small wrapper which contains `PackAuthoredDto` and `IndexStatusDto`. `PackAuthoredDto` contains the vehicles, saves, changelog, etc.

src/Borea.Storage/Index/SnapshotDto.cs
MrJeranimo: I added a `ContentIndexRootValidator` that validates the root and now `ContentIndexFetcher` now uses the shared root validator.

MrJeranimo:
I should have finished the DTOs. I also added a `ContentIndexRootValidator` and tests which will validate the basic root structure of the Index before deserialization. `ContentIndexFetcher` also now uses it before saving the file to have some reassurance it got a somewhat valid looking index before overwriting a saved index.

I will add the serializer in a new stacked PR so this and #96 can be merged as well.

src/Borea.Core/Index/ContentIndexRootValidator.cs
```diff
@@ -0,0 +1,136 @@
...
+        {
+            indexJson = JsonDocument.Parse(indexFile);
+        }
+        catch
+        {
+            throw new InvalidOperationException($"Index file is not valid JSON at {indexPath}");
+        }
+
+        return ValidateIndexRoot(indexJson);
+    }
```
MrJeranimo, line 72: I realize that this does not check if `indexFile` is empty or null. I will add that in my next stacked PR.

Maximilian-Nesslauer (approved):
LGTM, thank you very much for the PR :)
The current DTO contracts read the live snapshot fixture, and the shared root validator protects the fetched index. The remaining reader and mapping behavior can follow in the planned stacked pull requests.

PR #96, opened 2026-09-10
Add content index DTO boundary tests

Maximilian-Nesslauer:
## Summary

This adds the test-owned part of #44 on top of #89.

The tests cover the snapshot envelope and the DTO boundaries for listings, releases, and packs. They verify that a bad child document can be rejected without losing the outer entry data or the other child documents.

The test fixture contains the current content index data with four listings and ten releases. All listings and releases in this fixture deserialize with the current DTO contracts.

Part of #44.
Builds on #89.


averageksp (approved):
Love to see the new tests 😁

Issue #90, opened 2026-09-08
Persist Borea.App preferences and custom themes


Maximilian-Nesslauer:
# Persist Borea.App preferences and custom themes

## Current status

The #50 branch reads the selected theme and custom themes from JSON files under `src/Borea.App/client/`.

Those files are packaged with the app and are not a suitable location for user-writable data.

`BoreaSettings` and `IBoreaSettingsRepository` persist only operational settings, which are the game directory and installed loader directories.
They intentionally exclude app-level preferences because those settings do not affect the Core service graph.

## Scope

- Store the selected theme and custom themes under Borea's data directory.
- Keep the bundled default themes read-only.
- Load preferences without preventing the app window from opening when the preferences file is absent or invalid.
- Validate the saved theme selection and use a default theme when the named theme does not exist.
- Keep the preference model and persistence separate from `BoreaSettings`, the CLI, and game or loader configuration.

## Tests

- No saved preferences uses the default theme.
- A saved bundled theme is restored.
- A saved custom theme is restored.
- An invalid or unknown saved theme falls back to the default without preventing startup.

## Not in scope

- Game and mod-loader path configuration.

PlazmaBoltz:
Themes are currently stored in multiple JSON files at Borea.App/client. Default themes are stored in BoreaDefaultThemes.json and custom themes are stored in CustomThemes.json. Currently there is no way of editing themes in the UI (todo).

Theme data is stored as follows within the JSON format

```
{
     "Theme Name": [
          "Color1",
          "Color2",
          "AccentColor",
          "TextColor"
     ],
     "another theme"...
}
```

All colors are stored as HCC codes. No two themes  may have the same name since they are stored in a dictionary on load.

Current default themes are:
 - Borealis (Blue style, default)
 - Light Mode
 - Dark Mode

Any suggestions for default themes are appreciated.

Maximilian-Nesslauer:
> Themes are currently stored in multiple JSON files at Borea.App/client. Default themes are stored in BoreaDefaultThemes.json and custom themes are stored in CustomThemes.json. Currently there is no way of editing themes in the UI (todo).
> 
> Theme data is stored as follows within the JSON format
> 
> ```
> {
>      "Theme Name": [
>           "Color1",
>           "Color2",
>           "AccentColor",
>           "TextColor"
>      ],
>      "another theme"...
> }
> ```
> 
> All colors are stored as HCC codes. No two themes may have the same name since they are stored in a dictionary on load.
> 
[...]

PR #99, opened 2026-09-10
Add English and German localization to Borea.App

### LINKED ISSUE #95: Adding localization for languages
There are many people in the KSA Community who don't speak English and even more that don't have it as their native language. By adding localization, Borea will be more accessible to the community.

This will likely only be need for the UI because I do not expect many people to be using the app through command prompt.

Maximilian-Nesslauer:
## Summary

This adds ResX-based localization to the Borea desktop App. The current interface is available in English and German, and the language can change while the App is running.

The App uses the operating system UI language when it starts. An English or German regional culture, such as `en-US` or `de-DE`, selects its supported parent language. Any other or invalid language falls back to English.

Changing the UI language does not change the regional culture that formats dates, numbers, or other regional values. Only `CurrentUICulture` changes.

## User interface

The navigation labels, page headings, settings labels, logo placeholder, accessibility names, and missing-view message now come from the localization resources. The Settings page contains a language selector for English and German.

The UI binds to one observable localization service. When the user selects another language, the service notifies the active bindings and the visible text updates without an App restart. The missing-view fallback also listens for this notification while it is visible.

## Why this design

Avalonia documents ResX as its standard resource-based localization approach in [Localizing using ResX](https://docs.avaloniaui.net/docs/app-development/localizing). The guide also explains how the build generates a public, strongly typed `Resources` class. This change uses the cross-platform MSBuild generator described there, so the generated file stays under `obj` and does not depend on Visual Studio writing a source file into the repository.

.NET uses `CurrentUICulture` to select translated resources, while `CurrentCulture` controls regional formatting. The [.NET resources overview](https://learn.microsoft.com/en-us/dotnet/core/extensions/resources) explains this separation and how `ResourceManager` retrieves the correct resource set. Keeping these values separate lets a user select German interface text without unexpectedly changing number or date formatting.

The [.NET localization guide](https://learn.microsoft.com/en-us/dotnet/core/extensions/localization) explains the parent-culture and neutral-resource fallback rules. Borea first reduces a regional UI culture such as `de-DE` to the supported `de` language. It then uses the neutral English resource when no supported language matches, so the UI always has usable text.

The Avalonia guide shows static XAML resource references, but static references do not notify existing controls after a runtime language change. Borea instead exposes localized values through an observable service. Avalonia describes this update mechanism in [How to use INotifyPropertyChanged](https://docs.avaloniaui.net/docs/data-binding/inotifypropertychanged). One notification refreshes all localization bindings after the selected language changes.

The language and theme selectors also receive localized `AutomationProperties.Name` values. Avalonia explains in its [Accessibility guide](https://docs.avaloniaui.net/docs/app-development/accessibility) that these properties provide metadata to assistive technologies without changing the visual layout.

## Resource structure

The source resource files are under `Localization/Resources/`. The project generates one strongly typed `Resources` class during the build, so generated code is not committed.

The neutral resource contains English text. `Resources.de.resx` contains the German translation. A parity test requires both files to contain the same resource keys.

## Scope and preference storage

This change localizes `Borea.App` only. It does not change the CLI because issue #95 identifies the graphical interface as the required scope.

The selected language applies for the current App run. Persistent storage remains part of issue #90. This change exposes the canonical `SelectedCultureName` value and `TrySetCulture` method so the preference work can load and save `en` or `de` without duplicating localization rules.

Closes #95

### DIFF
--- a/Borea.slnx
+++ b/Borea.slnx
@@ -8,6 +8,7 @@
     <Project Path="src/Borea.Storage/Borea.Storage.csproj" />
   </Folder>
   <Folder Name="/tests/">
+    <Project Path="tests/Borea.App.Tests/Borea.App.Tests.csproj" />
     <Project Path="tests/Borea.Cli.Tests/Borea.Cli.Tests.csproj" />
     <Project Path="tests/Borea.Composition.Tests/Borea.Composition.Tests.csproj" />
     <Project Path="tests/Borea.Core.Tests/Borea.Core.Tests.csproj" />
--- a/README.md
+++ b/README.md
@@ -4,6 +4,9 @@ Borea is a cross-platform complete general content manager for Kitten Space Agen
 # Credits
 - [MrJeranimo](https://github.com/MrJeranimo) - Original Creator and Developer
 
+# Contributing translations
+You do not need to write C# to improve an existing translation. See the [localization guide](docs/localization.md) for instructions to correct text or propose a new language.
+
 # Repository Structure
 | Path | Description |
 | --- | --- |
--- a/docs/localization.md
+++ b/docs/localization.md
@@ -1,0 +1,89 @@
+# Help translate Borea
+
+You do not need to write C# to improve an existing Borea translation. You can edit a resource file through the GitHub website and submit the change for review.
+
+## Find the language files
+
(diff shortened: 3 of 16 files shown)


averageksp (approved):
Lgtm

PR #108, opened 2026-09-12
Added Index -> DTO deserializers and validators. Added DTO -> `Borea.Core` objects mappers.

MrJeranimo:
This is the second part of the index-validator, the parsers, validators, and mappers. This PR adds the Index -> DTO deserializers and validators, and DTO -> `Borea.Core` objects mappers. The mappers are all done through `DtoMapper` as only the file to convert the index to `Borea.Core` objects should call it; this can be changed.

Also adds two new commands to `Borea.Cli` for testing the validators and mappers on the actual index.

Moves all the index DTOs into `Borea.Storage.Index.Dtos` folder for better organization (there would be about 40 files in `Borea.Storage.Index` otherwise.

This PR currently does not include a class that will automatically move the valid `ModMetadata`, `ModVersionMetadata`, or `ModPackMetadata` into the respective repositories.


src/Borea.Storage/Index/ListingParser.cs
```diff
@@ -0,0 +1,76 @@
...
+                    unknownReleases.Add(outcome.Unknown!);
+                    break;
+                case ParseOutcomeKind.Malformed:
+                    rejectedReleases.Add(outcome.Malformed!);
+                    break;
+            }
+        }
+
+        return ParseOutcome<ParsedListing>.Valid(new ParsedListing(
+            listing.Id, listing.Authored, validReleases, rejectedReleases, unknownReleases, listing.IndexStatus));
```
Maximilian-Nesslauer, line 74: Follow-up: keep constructor failures inside the entry boundary. Setting one listing's outer `id` to whitespace makes `new ParsedListing(...)` throw `ArgumentException`, which escapes `SnapshotParser.Parse` and prevents every other entry from being returned. `PackParser` has the same problem, including its tombstone path. Please validate the ID before construction, or convert expected constructor failures into a malformed entry, and test a bad ID beside a valid sibling.

src/Borea.Cli/Commands/IndexCommand.cs
```diff
@@ -25,4 +30,94 @@ private static Command BuildRefresh(Func<CancellationToken, Task<CliServices>> s
...
+                if (listing.Authored is not null)
+                    authored = DtoMapper.MapAuthored(listing.Authored, "index");
+
+                IndexStatus? indexStatus = null;
+                if (listing.IndexStatus is not null)
+                    indexStatus = DtoMapper.MapIndexStatus(listing.IndexStatus);
+
+                foreach (var release in listing.ValidReleases)
+                {
+                    DtoMapper.MapRelease(release, "index", authored);
```
Maximilian-Nesslauer, line 99: Follow-up before repository integration:

mapping needs the same per-entry failure boundary as deserialization, as discussed in #89. A release with `version: "invalid"` or `download.size: -1` enters `ValidReleases`, then throws here and stops processing the remaining entries. A null dependency item also passes deserialization and throws during mapping. Please combine semantic validation and mapping into an entry result that keeps the ID/version and reason, while continuing with valid siblings. The authored and pack-version paths need the same treatment; a whole-command catch does not provide that isolation.

src/Borea.Storage/Index/DtoMapper.cs
```diff
@@ -0,0 +1,334 @@
...
+
+    private static LoaderProvides MapProvides(ProvidesDto dto) =>
+        new(
+            launch: dto.Launch,
+            contentDir: dto.ContentDirectory is null ? null : MapInstallAnchor(dto.ContentDirectory),
+            contentPath: dto.ContentPath,
+            configure: dto.Configure is null ? null : MapConfigure(dto.Configure));
+
+    private static LoaderConfigure MapConfigure(ConfigureDto dto) =>
+        new(dto.File, MapConfigureFormat(dto.Format), dto.GamePath);
```
Maximilian-Nesslauer, line 282: Follow-up before loader integration:

this drops `ConfigureDto.UnknownFields`. Adding an unknown key to StarMap's `provides.configure` still produces a successful `LoaderConfigure`, with the extra key lost. RFC 0035 deliberately makes this table closed, and #89 kept extension data so the mapper could reject it. Please reject a nonempty `UnknownFields` collection and retain the listing ID and reason at the entry boundary, so we do not configure a loader after silently omitting part of its instructions.

src/Borea.Storage/Index/DtoMapper.cs
```diff
@@ -0,0 +1,334 @@
...
+    private static ListingSnapshot? MapListingSnapshot(JsonElement element, ModMetadata? authored)
+    {
+        ReleaseListingDto? dto;
+        try
+        {
+            dto = element.Deserialize<ReleaseListingDto>(IndexJsonOptions.Value);
+        }
+        catch (JsonException)
+        {
+            dto = null;
```
Maximilian-Nesslauer, line 303: Follow-up: only absent frozen fields should use the live listing as a fallback.

With `listing: { "name": "Frozen name", "authors": 42 }`, this catch discards the valid frozen name too, and mapping succeeds with the current authored name. #89 explicitly agreed that a present field of the wrong type makes that release unreadable. Please report that release as malformed, keep its valid siblings, and retain field-by-field fallback for absent fields. The malformed-listing and empty-authors tests currently expect the opposite and need to change with the fix.

src/Borea.Storage/Index/ListingParser.cs
```diff
@@ -0,0 +1,76 @@
...
+public static class ListingParser
+{
+    public static ParseOutcome<ParsedListing> Parse(JsonElement element)
+    {
+        var id = IndexJsonHelpers.TryExtractString(element, "id");
+
+        ListingEntryDto listing;
+        try
+        {
+            listing = element.Deserialize<ListingEntryDto>(IndexJsonOptions.Value)
```
Maximilian-Nesslauer, line 20: Follow-up: read `authored.spec_version` from the raw element before deserializing the v1 DTO, as the release and pack-version parsers already do. A listing with `spec_version: 2` and a changed `name` shape, such as an object, currently becomes malformed before reaching the version check. It should remain an unknown newer-format entry with its ID and spec version. The current test only changes the version number while leaving the v1 shape intact.

src/Borea.Storage/Index/ListingParser.cs
```diff
@@ -0,0 +1,76 @@
...
+                case ParseOutcomeKind.Unknown:
+                    unknownReleases.Add(outcome.Unknown!);
+                    break;
+                case ParseOutcomeKind.Malformed:
+                    rejectedReleases.Add(outcome.Malformed!);
+                    break;
+            }
+        }
+
+        return ParseOutcome<ParsedListing>.Valid(new ParsedListing(
```
Maximilian-Nesslauer, line 73: Follow-up before repository integration: validate the outer ID with ModIds and require it to agree, case-insensitively, with authored.id and each release ID. The parser currently accepts a wrapper ID such as "My Mod!", and a listing named Probe can carry a release whose mapped ModId is StarMap. This does not demonstrate an overwrite: FileModInstaller refuses an existing installed ID or foreign folder. It does mean the reader can associate a selected listing with the wrong release identity. Please also detect duplicate IDs across listings and packs, using the shared case-insensitive comparer.

src/Borea.Storage/Index/DtoMapper.cs
```diff
@@ -0,0 +1,334 @@
...
+        _ => throw new FormatException(
+            $"Host reference must be a string or number, but was {value.ValueKind}."),
+    };
+
+    private static ModDependency MapDependency(DependencyEntryDto dto)
+    {
+        var kind = MapDependencyKind(dto.Kind);
+        var source = dto.Source is null ? null : (MetadataSource?)MapMetadataSource(dto.Source);
+
+        if (dto.AnyOf is { Count: > 0 })
```
Maximilian-Nesslauer, line 231: Follow-up: validate dependency shape and collection elements before mapping. When both id and a nonempty any_of are present, this branch silently discards id; when any_of is empty, it falls through to the single-ID path. ModDependencyMapper already has explicit checks for these shapes, although its persistence DTO prevents calling it directly here. Share or restate those checks and report failures at the entry boundary. Core already rejects an invalid any_of kind. Also reject null collection elements: authors: ["A", null] currently reaches ModMetadata, while dependencies: [null] fails later with NullReferenceException. Add tests beside valid siblings.

src/Borea.Storage/Index/DtoMapper.cs
```diff
@@ -0,0 +1,334 @@
...
+        if (dto.OtherLinks is not null)
+        {
+            foreach (var (key, value) in dto.OtherLinks)
+            {
+                // Extension-data entries are arbitrary JSON; documented
+                // links are plain URL strings, so anything else is skipped
+                // rather than failing the whole listing over an unrelated field.
+                if (value.ValueKind == JsonValueKind.String)
+                {
+                    links[key] = value.GetString()!;
```
Maximilian-Nesslauer, line 199: Follow-up: prevent extension link keys from replacing a declared key. MapLinks inserts dto.Forums first, then writes extension data into an OrdinalIgnoreCase dictionary. A document with both "forums" and "Forums" therefore replaces the declared forums URL with the extension value. Please reject a case-insensitive collision, consistent with the Core model's duplicate-key check, and report the listing ID and reason through the entry boundary.

Maximilian-Nesslauer (approved):
LGTM 👍 :)

This is a useful next step for #44. The DTO boundaries, shared root validator, and test coverage give us a good base for repository integration.

I am happy to merge this step and take those fixes in a follow-up before the mapped data feeds the repositories and installer.

PR #110, opened 2026-09-12
Complete the content index reader and repository

### LINKED ISSUE #44: Read the index snapshot as a mod source

Maximilian-Nesslauer:
Closes #44

Borea now reads the content index into its mod repository and uses it before fallback sources.
Invalid entries produce diagnostics while valid entries remain available, and a failed refresh keeps the last working cache.
The CLI reports accepted content and detailed diagnostics as text or JSON.


MrJeranimo (approved):
The parsers, validators, and mappers seem much more robust now. Also all tests and the CLI commands work on my machine.

PR #111, opened 2026-09-12
Update loader configuration when the game path changes

### LINKED ISSUE #104: Update installed loader configuration after the game path changes
## Implementation

The implementation is proposed in #111.

# Update installed loader configuration after the game path changes

## Simple explanation

A loader needs to know where the game is installed.
If a user changes Borea's game path, an installed loader can still point at the old game directory.
The next launch can then start the wrong game, or fail because that old directory no longer exists.

## Current status

`WriteGamePathAsync` runs when `FileLoaderInstaller` installs a loader.
Changing the game path through Borea settings does not update existing loader configurations.

## Goal

When Borea changes its configured game path, update the configured game path for every installed loader that declares `[provides.configure].game-path`.
Keep the settings change and the loader updates consistent: Borea must not report the new game path if a required loader configuration update fails.

## Today and should be

| Today | Should be |
|---|---|
| A loader can retain a stale game path after a settings change. | Each configured loader follows the selected game path. |
| Launch can use a removed or unintended game installation. | Launch uses the installation selected in Borea settings. |

## Tests

- Change the game path with one configured loader and verify its configuration file changes.
- Verify that loaders without a game-path configuration declaration are left unchanged.
- Verify the result and stored settings when an update fails.

## Not in scope

- Adopting a loader.
- Installing or removing a loader.
- Per-instance game paths.

## Depends on

No open issue blocks this work.

Maximilian-Nesslauer:
Closes #104

Changing the game directory now updates installed loaders that declare a game-path configuration key.
Borea saves the new setting after those updates succeed and restores the prior files and settings when an update fails.
Rollback failures are reported explicitly.

Stacked on #110.

### DIFF
--- a/src/Borea.Cli/CliServices.cs
+++ b/src/Borea.Cli/CliServices.cs
@@ -22,6 +22,8 @@ internal sealed class CliServices : IDisposable
 
     public required IBoreaSettingsRepository SettingsRepository { get; init; }
 
+    public required IGameDirectoryChanger GameDirectoryChanger { get; init; }
+
     public required IInstanceRepository Instances { get; init; }
 
     public required IModStateRepository ModState { get; init; }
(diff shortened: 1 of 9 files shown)


MrJeranimo (approved):
LGTM

PR #112, opened 2026-09-12
Add loader management and instance launch commands

### LINKED ISSUE #103: CLI slice 4: manage loaders and launch instances
## Implementation

The implementation is proposed in #112.

# CLI slice 4: manage loaders and launch instances

## Simple explanation

Borea can install and adopt a loader in its services, but a person cannot ask Borea to do this from the command line.
This task connects the completed loader functions to the CLI so a person can install, adopt, remove, and use a loader.

## Current status

`ILoaderInstaller`, `ILoaderAdopter`, `ILoaderUninstaller`, and the loader launch service exist on `main`.
The CLI has no commands that call the three loader services.
`LoaderLauncher` is also absent from `BoreaServices`, so no composition root provides it.
`settings set loader <id> <directory>` creates a `LoaderInstallation` directly instead of using adoption.

## Goal

Add CLI commands that let a user install, adopt, uninstall, and launch a mod loader.
Register `LoaderLauncher` in `BoreaServices`.
Change `settings set loader` to adopt the selected directory through `ILoaderAdopter`, so Borea validates the executable, reads its version, and verifies the configured game path before it stores the installation.

## Today and should be

| Today | Should be |
|---|---|
| A user cannot manage a loader from the CLI. | A user can install, adopt, and uninstall a loader from the CLI. |
| A user cannot launch an instance through the loader. | A user can launch an instance when its loader declares the required handover. |
| `settings set loader` records unverified data. | The command uses the same adoption validation as every other adoption path. |

## Tests

- Add CLI tests for each command and its main failure result.
- Verify that adoption rejects a missing executable and reports warnings for an unreadable version or a mismatched game path, as defined by the merged adoption service.
- Verify that a configured instance launches through the registered launcher.

## Not in scope

- New loader metadata fields.
- Borea.App screens.
- Changing the instance handover format. That remains blocked by design RFC 0049.

## Depends on

The task can use the merged loader services and launcher work.
It is a child of #65 and does not require Borea #44.

Maximilian-Nesslauer:
Closes #103

The CLI can now install, adopt and uninstall loaders, and launch an instance through an installed loader.
The loader setting uses the existing adoption service, including its warnings for an unreadable version and a different configured game path.
Explicit yanked releases are refused before installation, and cancellation is checked before process startup.

Stacked on #111.

### DIFF
--- a/src/Borea.Cli/BoreaCli.cs
+++ b/src/Borea.Cli/BoreaCli.cs
@@ -26,6 +26,7 @@ internal static class BoreaCli
         root.Subcommands.Add(ModStateCommands.BuildEnable(services));
         root.Subcommands.Add(ModStateCommands.BuildDisable(services));
         root.Subcommands.Add(LoaderCommand.Build(services));
+        root.Subcommands.Add(LaunchCommand.Build(services));
         return root;
     }
 
--- a/src/Borea.Cli/BoreaCli.cs
+++ b/src/Borea.Cli/BoreaCli.cs
@@ -25,6 +25,7 @@ internal static class BoreaCli
         root.Subcommands.Add(InstanceCommand.Build(services));
         root.Subcommands.Add(ModStateCommands.BuildEnable(services));
         root.Subcommands.Add(ModStateCommands.BuildDisable(services));
+        root.Subcommands.Add(LoaderCommand.Build(services));
         return root;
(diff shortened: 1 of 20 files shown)


MrJeranimo (approved):
Absolutely crushing it Maxi

PR #116, opened 2026-09-12
Read mod packs from the content index

### LINKED ISSUE #115: Read mod packs from the cached content index
## Implementation

The implementation is proposed in #116.

## Current status

IModPackRepository defines pack queries but has no implementation.
The pack metadata model and parsers already exist.
PR #110 reads packs into ContentIndexSnapshot.Packs, including pack and version status, but connects only the mod repository to the production services.

The contract is [spec/snapshot.md](https://github.com/KSAModding/content-manager-design/blob/main/spec/snapshot.md).

## Goal

Implement IModPackRepository using the same snapshot and cache as the mod index source, and register it in the production services.
Pack queries must not add a second fetcher, cache file or refresh policy.

| Today | Should be |
| --- | --- |
| Mapped packs have no repository consumer. | Callers can list and search packs, list versions, and retrieve the latest or an exact version. |
| Repository methods return metadata without index status. | Callers can obtain pack and version status and diagnostics without reading Storage DTOs. |
| The interface describes only available, non-removed versions. | Normal latest selection skips retracted versions, while an explicit version lookup retains the retracted version and its warning. |

Use the existing mod version ordering and case-insensitive ID rules.
Define ordering for pack lists, search results and version lists so repeated queries are stable.
Delisted packs do not produce usable metadata, but their tombstone and identity remain available through diagnostics or an explicit result.
Disputed entries and unknown status values remain visible with warnings.
An unsupported document version retains its identity and diagnostic state and does not become usable metadata.
Supported sibling versions remain available.
Adjust the Core query contract where necessary so callers can distinguish unknown or delisted content from an ID that was never present.
Reuse the existing diagnostic types where they fit.

## Tests

Cover search and no match, case-insensitive lookup, version ordering, and latest selection when the newest version is retracted.
Cover explicit retrieval of a retracted version, disputed packs, delisted tombstones, unknown status values and unsupported document versions beside supported siblings.
Verify that mod and pack queries share refresh and cache behavior, that offline queries use the cached snapshot, and that a failed refresh keeps the previous usable snapshot.
Verify production service registration.

## Not in scope

Installing pack members, changing index publishing, adding CLI commands or App screens.
Those installation services remain #7.

## Depends on

#44, implemented by PR #110.
This issue is a prerequisite of #7.

Maximilian-Nesslauer:
Review order: this repository change comes before https://github.com/KSAModding/Borea/pull/117, which uses its shared snapshot provider.

Closes #115.

Borea could parse pack metadata, but callers had no repository through which to find packs or select a version.
This adds those queries and returns the pack's moderation status and diagnostics with the result.
It shares the mod repository's snapshot and cache, so mod and pack queries do not download the index separately.

For example, if pack version 2.0.0 is retracted and 1.0.0 is available, normal latest selection returns 1.0.0.
An explicit request for 2.0.0 still returns that version with its retracted status, allowing a caller to show the warning.
Delisted or unsupported content retains its identity without becoming usable pack metadata.

This supplies pack discovery and version selection.
Installing the selected pack's members remains #7.

### DIFF
--- a/src/Borea.Composition/BoreaServices.cs
+++ b/src/Borea.Composition/BoreaServices.cs
@@ -85,6 +85,8 @@ public sealed class BoreaServices : IDisposable
     /// </summary>
     public required IModRepository Mods { get; init; }
 
+    public required IModPackRepository ModPacks { get; init; }
+
     public required IModDownloader Downloader { get; init; }
 
     public required ILoaderInstaller LoaderInstaller { get; init; }
@@ -102,6 +104,8 @@ public sealed class BoreaServices : IDisposable
     public required IContentIndexFetcher IndexFetcher { get; init; }
 
     public required IContentIndexReader IndexReader { get; init; }
+
+    public required IContentIndexSnapshotProvider IndexSnapshots { get; init; }
 
     public required IContentIndexRepository ContentIndex { get; init; }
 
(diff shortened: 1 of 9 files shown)


MrJeranimo (approved):
LGTM
(I couldn't help myself, I like working on the project)

PR #119, opened 2026-09-12
Add a command `loader list` to the CLI. 

MrJeranimo:
This command lists the installable mod loaders. There is an optional flag `-v` or `--versions` that will also print out all the installable versions of the mod loaders.


Maximilian-Nesslauer (approved):
LGTM, thank you for the PR :)

The loader list command looks good to merge.
I checked both version flags, version ordering, yanked-release filtering, an empty catalogue and repository failure handling.

I will add focused command tests in a follow-up PR because the added root-help assertions do not exercise loader list itself.

PR #145, opened 2026-09-13
Add an About tab to the settings

averageksp:
## Add an About tab to the settings

Settings gets a third tab next to General and Game.

### What it shows
- Borea version (from the version the release workflow stamps; a local build says 1.0.0), the .NET runtime, the operating system, and the installed game build.
- **Copy diagnostics**: puts those lines plus the recorded mod loaders on the clipboard, for pasting into a bug report.
- **Folders**: the Borea folder path, with buttons to open it and the instances folder.
- **Links**: source code, report a bug, releases.
- **Third-party notices**: Avalonia, CommunityToolkit.Mvvm, IBM Plex and Phosphor Icons with their licenses. OFL and MIT both ask for this and nothing showed it before.

### How
- The settings tab state is an enum now (`General`, `Game`, `About`) instead of one bool, so a third tab fits.
- Folders and links open through the system, the same way links on a content page do. A folder that doesn't exist yet reports that instead of failing.
- The clipboard needs the window, so the copy button is handled in `MainWindow.axaml.cs` with the text coming from the view model.
- No Core or Storage changes.

### Not in this PR
- A log file and an "Open logs" button. Borea writes no log file yet, so that needs its own small piece of infrastructure first.
- Update checking.

### Tests
Six new view model tests: tab switching, version and environment values, folder paths, diagnostics text, and the missing-folder and bad-link cases. All test projects pass, `dotnet format` is clean.


src/Borea.App/ViewModels/MainViewModel.About.cs
```diff
@@ -0,0 +1,147 @@
...
+
+    public static string RuntimeText { get; } = RuntimeInformation.FrameworkDescription;
+
+    public static string SystemText { get; } = $"{RuntimeInformation.OSDescription} ({RuntimeInformation.OSArchitecture})";
+
+    /// <summary>
+    /// The work of others that ships inside Borea, with the license each one
+    /// requires to be named.
+    /// </summary>
+    public static IReadOnlyList<ThirdPartyNotice> ThirdPartyNotices { get; } =
```
Maximilian-Nesslauer, line 45: [should fix] This lists 4 of the 31 packages in `src/Borea.App/packages.lock.json`, and Tomlyn (BSD-2-Clause) is missing completely, as are SkiaSharp, HarfBuzzSharp, the ANGLE natives, Tmds.DBus, MicroCom and the Microsoft.Extensions packages, plus the .NET runtime in the self-contained build. MIT and BSD also ask for the copyright and permission notice text, which a name with a license label does not carry. I would word this as a short credits list, and as a follow-up have the release workflow write a THIRD-PARTY-NOTICES file into each archive from the lock files (the SBOM already has every license) and let About open it.

src/Borea.App/ViewModels/MainViewModel.About.cs
```diff
@@ -0,0 +1,147 @@
...
+                .AppendLine(SystemText)
+                .AppendLine($"KSA: {InstalledVersionText ?? "not set up"}");
+
+            if (_services is not null)
+            {
+                // the matched release is what the Game tab shows; the file version is the fallback for an adopted loader
+                foreach (var (loaderId, installation) in _services.Settings.LoaderInstallations)
+                {
+                    var version = installation.Version?.ToString() ?? installation.RawVersion ?? "version unknown";
+                    text.AppendLine($"{loaderId}: {version} at {installation.DirectoryPath}");
```
Maximilian-Nesslauer, line 85: [should fix] This line carries the user name through the loader directory path, for example `C:\Users\<name>\AppData\Local\Borea\Loaders\StarMap`, in a text meant for public bug reports. Could we shorten the user profile prefix to `~` before it goes on the clipboard?

src/Borea.App/ViewModels/MainViewModel.About.cs
```diff
@@ -0,0 +1,147 @@
...
+        if (string.IsNullOrWhiteSpace(target))
+            return;
+
+        AboutMessage = null;
+        try
+        {
+            if (Directory.Exists(target) || Uri.IsWellFormedUriString(target, UriKind.Absolute))
+                Process.Start(new ProcessStartInfo(target) { UseShellExecute = true });
+            else
+                AboutError = Localization.FormatAboutFolderMissing(target);
```
Maximilian-Nesslauer, line 133: [minor] A link that is not a well-formed URL also lands here, so the message would say "not a link does not exist yet." Only constants reach this today, so it is only the wording, but a separate "cannot open" message would be clearer.

src/Borea.App/ViewModels/MainViewModel.About.cs
```diff
@@ -0,0 +1,147 @@
...
+        {
+            AboutError = exception.Message;
+        }
+    }
+
+    private static string ReadVersion()
+    {
+        var informational = typeof(MainViewModel).Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion;
+        if (!string.IsNullOrWhiteSpace(informational))
+            return informational.Split('+')[0];
```
Maximilian-Nesslauer, line 145: [should fix] The dropped part is the commit hash: a local build of this branch reports `1.0.0+c47bff52...`. Fine for the version row, but the diagnostics should keep the full value, otherwise every local build reports `Borea 1.0.0` and two builds cannot be told apart.

tests/Borea.App.Tests/ViewModels/AboutViewModelTests.cs
```diff
@@ -0,0 +1,106 @@
...
+    [Fact]
+    public async Task OpenLink_NotAUrlOrFolder_ReportsInsteadOfStarting()
+    {
+        using var harness = await ViewModelHarness.CreateAsync();
+        var viewModel = harness.ViewModel;
+
+        viewModel.OpenAboutLinkCommand.Execute("not a link");
+        Assert.NotNull(viewModel.AboutError);
+
+        viewModel.OpenAboutLinkCommand.Execute(null);
```
Maximilian-Nesslauer, line 113: [minor] `OpenFromAbout` returns early for null, so the next `Assert.NotNull` only passes because the error from the line before is still set. Assert `Null` after resetting `AboutError`, or drop the null case.

Maximilian-Nesslauer (commented):
Nice addition :) LGTM but i posted some inline comments that should be done in an follow up PR.

1. [should fix] The third-party list is much shorter than what ships. The PR text says OFL and MIT ask for this, but the App archive contains 31 NuGet packages, not 4. Tomlyn is BSD-2-Clause and is missing completely, and so are SkiaSharp, HarfBuzzSharp, the ANGLE natives, Tmds.DBus, MicroCom and the Microsoft.Extensions packages, plus the .NET runtime itself in the self-contained build. MIT and BSD also ask for the copyright and permission notice text, and a name with a license label does not carry that. I would either word the section as a short credits list, or, better as a follow-up, have the release workflow write a THIRD-PARTY-NOTICES file into each archive from the lock files (the SBOM already has every license) and let About open it.
2. [should fix] "Copy diagnostics" puts the full loader path into a text meant for public bug reports, for example `C:\Users\<name>\AppData\Local\Borea\Loaders\StarMap`. That publishes the Windows user name. Replacing the user profile prefix with `~` (or `%USERPROFILE%`) keeps the report useful.
3. [should fix] `ReadVersion` drops the part after `+`. That part is the commit hash (a local build of this branch reports `1.0.0+c47bff52...`), and it is the only thing that tells two local builds apart, because every local build says `1.0.0`. I would keep the short version on the page and put the full informational version into the diagnostics.

Inline comments below for the smaller items.

Maximilian-Nesslauer (approved):

PR #158, opened 2026-09-13
Prompt for the game folder when none is saved

### LINKED ISSUE #152: Prompt for setup when no game directory is saved
## Prompt for setup when no game directory is saved

When nothing is configured, Home shows a "Game not set up" chip and that is it. New users don't know that Settings → Game is where they need to go, and they only hit the problem when an install fails to plan.

### Proposal

- On startup with no game directory, open Settings on the Game tab, or show a banner on Home ("Borea needs to know where KSA is") with a button that opens it.
- Show the same banner on Discover and Library while the game is not set up, since installs and Play are the actions that need it.
- Once the game directory is saved, the prompt goes away and does not come back unless the folder is missing.

What could also work: If auto-detection finds the game, the prompt can offer "Use this folder" instead of sending the user to the picker.

averageksp:
## Prompt for the game folder when none is saved

Closes #152.

Without a game directory, Home showed a "Game not set up" chip and nothing else. Now:

- The first load without a saved game directory opens Settings on the Game tab, once per session.
- Home, Discover and Library show a banner under the heading with a "Set up game" button that opens the Game tab. When a saved folder no longer exists, the banner says that instead and the settings do not open on their own.
- Saving a valid folder removes the banner.

The banner is one `DataTemplate` used on the three pages. `GameSetupState` on the view model drives it and is refreshed on every load, so a rebuild after saving updates it.


Maximilian-Nesslauer (commented):
lgtm but needs conflicts resolved

averageksp:
Rebased

Maximilian-Nesslauer (approved):

PR #159, opened 2026-09-13
Draw the Discover add button as a purple plus on grey

### LINKED ISSUE #154: Discover: install button should be the purple plus on a grey card
## Discover: install button should be the purple plus on a grey card

The "+" install button on the Discover rows is a white plus on a black circle. In the design it is a purple plus on a grey circle, and the whole row sits on a grey card, not black.

### Current

- `Button.add` in `App.axaml` uses `Brush.SurfaceHeader` (black) as background and the plus is drawn with `Brush.Text` (white).
- The rows around it use the dark header colour too, so the button does not stand out from its card.

### Expected (design)

- Circle: the grey card surface (`Brush.Surface`), a slightly lighter grey on hover.
- Plus: `Brush.Accent` (purple), 16 px, like the plus already used in the Versions table on the content page.
- Row card: grey surface with the usual card radius, the black header colour stays for the top bar of the Current Install card only.

Screenshot from the design attached for reference. The content page already draws the plus in accent on the version rows, so this is aligning Discover with that.

Image

averageksp:
## Draw the Discover add button as a purple plus on grey

Closes #154.

`Button.add` used the black header colour with a white plus, so it did not stand out from the grey row card. It now uses the raised grey surface with the accent plus, like the plus on the version rows of the content page. The hover shade is unchanged.

Light theme picks up the same tokens, so nothing changes there beyond the plus being purple.


Maximilian-Nesslauer (approved):
lgtm

PR #165, opened 2026-09-13
Remove an installed mod from its Discover row

### LINKED ISSUE #164: Discover: uninstall a mod from its row
## Discover: uninstall a mod from its row

Once a mod is installed, its Discover row only shows an "Installed" chip. To remove it you have to go to Library, open the active instance, find the mod in the content list and remove it there. Installing took one click on the same page, so removing should too.

### Proposal

- On an installed row, replace the static "Installed" chip with a small menu (the same "⋮" as the instance rows) or turn the chip into a button with **Remove** as the action.
- Removing asks for confirmation inline, the way the instance content page does (`BeginRemove` / `ConfirmRemove` on `ContentItem`), and reuses `RemoveContentAsync`. No new backend code: `ModDependencyResolver.CheckUninstall` already refuses when another installed mod requires it, and that message can show on the row like the install errors do.
- After removal the row goes back to the "+" state and the instance mod count updates, since `ReloadInstancesAsync` already runs after a change.
- Removal only applies to the active instance, which is also what the "+" installs into. The row's tooltip should say which instance it acts on, so the user is not surprised when the mod is still in another instance.

### Also

The content page (open a mod, header button) has the same gap: it shows "Add" but nothing for an installed mod. The same Remove action belongs there.

### Out of scope

Mods Borea does not own (foreign installs) stay as they are; those are not removable from the library page today either.

averageksp:
## Remove an installed mod from its Discover row

Closes #164.

An installed row on Discover only showed the "Installed" chip, removing meant going through Library. Now:

- The chip gets a "⋮" menu with Remove. Confirming inline (same texts and buttons as the instance page) removes the mod from the active instance, and the row returns to "+".
- The chip's tooltip names the instance the row acts on ("Installed in Main"), since the same mod may sit in another instance.
- Foreign files and mods another mod requires stay, with the reason on the row, through the same `TryRemoveContentAsync` the instance page uses. No new backend code.
- The content page gets the same action. Its "⋮" sits next to the Add button as the design draws it, is always there for mods, and Remove in it is enabled once the mod is installed.

Fixes that showed up while testing:

- Installing or removing a mod sent you to the instance page if you had opened that instance earlier. The reload after a change now refreshes the instance page only when you are on it, so Discover and the content page stay where they are.
- An install error stayed on the content page after leaving it and was shown twice (under the abstract and above the tabs). Leaving the page or opening another mod now clears the message, a pending "install anyway" plan and the remove confirmation, on the header and on the version rows. It is shown once.
- An error on a version row was placed in the 80 px action column and overflowed the table. It wraps on its own line under the row now, like the "install anyway" warning.

Tests: remove after confirmation, removing from Discover stays on Discover, foreign files refuse with the reason, no active instance does nothing, leaving the page forgets the last outcome.


Maximilian-Nesslauer (commented):
conflicts need to be resolved

Maximilian-Nesslauer (approved):

PR #168, opened 2026-09-13
Split MainWindow into page, modal and settings controls

averageksp:
## Split MainWindow into page, modal and settings controls

`MainWindow.axaml` had grown to 1011 lines with every page, both modals and the settings tabs in one file. It is now split without changing any markup:

- `Views/MainWindow.axaml`: the nav rail, the page hosts, the error bar and the modal hosts (90 lines).
- `Views/Pages/`: `HomePage`, `DiscoverPage`, `LibraryPage`, `InstancePage`, `ContentPage`, `TasksPage`.
- `Views/Settings/`: `SettingsModal` with `GeneralSettings`, `GameSettings` and `AboutSettings`.
- `Views/NewInstanceModal.axaml`.
- `Views/Templates.axaml`: the row templates (`InstanceRow`, `ContentRow`, `DiscoverRow`, `GameSetupBanner`), merged into the application resources so every page can use them.

Each control declares `x:DataType="vm:MainViewModel"` and inherits the window's DataContext, so the bindings are unchanged. Only `Grid.Column` and `IsVisible` moved from each section's root onto its host element.

The click handlers moved with their controls: copying diagnostics and the Borea folder into `AboutSettings`, the folder pickers into `GameSettings`. They reach the clipboard and the storage provider through `TopLevel.GetTopLevel(this)` instead of the window.

No behaviour change. Please avoid other XAML PRs until this is merged, since any of them would conflict.


Maximilian-Nesslauer (approved):
lgtm

PR #171, opened 2026-09-13
Show the install step, size and time left in the App

### LINKED ISSUE #148: Show what an install is doing and how long it will take
## Show what an install is doing and how long it will take

During a mod or loader install the app shows a percentage bar and nothing else. Installs go through several steps (planning, downloading, verifying the checksum, extracting, writing the manifest), and for large downloads the user has no idea how long they will wait.

### What exists

`DownloadProgress(BytesDownloaded, TotalBytes)` is reported by `HttpModDownloader` and passed through `IModInstaller`, `IModReplacer`, `ILoaderInstaller` and `InstallPlanExecutor`. Only the download step reports, and only bytes.

### Proposal

- Widen the progress type to carry a phase (`Planning`, `Downloading`, `Verifying`, `Extracting`, `Finishing`) and the mod id and version being handled, so a plan with dependencies can say "Downloading MeasureTools 1.1.10 (2 of 3)".
- Have the installer report the phase changes, not just the download bytes.
- In the App, show the phase text under the bar, the size as "12.4 of 38.0 MB", and an estimate from the download rate over the last few seconds ("about 2 minutes left"). Skip the estimate for the non-download phases, they are short.
- The CLI can print the same phase lines.

Keep the estimate honest: it should say "about" and disappear when the rate is unknown.

averageksp:
## Show the install step, size and time left in the App

Closes #148. Builds on #170 

While a mod or loader installs, the App now shows under the progress bar:

- what is happening: "Downloading MeasureTools 1.1.10 (2 of 3)", "Unpacking …", "Configuring …", "Finishing …"
- while downloading: "12.4 of 38.0 MB, about 2 min left"

The time left comes from the download rate over the last five seconds. It appears after one second of data, is rounded to 5 seconds or to whole minutes, and is left out when the size or the rate is unknown. Each operation of a plan starts its own rate.

Shown on the Discover row, the content page header, the version rows and the loader install in Settings -> Game. English and German strings.

`InstallProgressText` formats the reports and takes a `TimeProvider`; tests cover the status and step, the size, seconds and minutes left, an unknown size, a new operation resetting the rate, and the German wording.


Maximilian-Nesslauer (changes requested):
lgtm but  "MB" is really MiB. The size is divided by 1024 × 1024 but labeled "MB". also  "Inconsistent visibility binding. In Settings, the status line shows when SetupProgressStatus is set. On the rows it shows when IsInstalling is true. Both work."

Maximilian-Nesslauer (approved):

PR #172, opened 2026-09-13
Notice a KSA update when the window is activated

### LINKED ISSUE #169: App: notice a KSA update while Borea is open
## App: notice a KSA update while Borea is open

When a new KSA release is installed while the App is running, parts of the App keep showing the old game version until Borea restarts or the game folder is saved again.

### What happens

`InstalledGameVersionProvider` reads `KSA.dll` on every call and caches nothing, so the CLI and the App's install planner always see the current build. The App only asks at certain moments, though:

- The "KSA …" chip on Home and in Settings → About is read in `LoadAsync`, which runs at startup and after a settings save.
- The Compatible / Untested / Incompatible chips on Discover, and the "Hide incompatible" filter, are computed once in `RefreshCompatibilityAsync` when Discover first loads.

After updating the game, the chip shows the old build, and Discover can say "Compatible" for a mod that the planner then blocks as incompatible, or the other way round.

### Proposal

- Read the installed version again when the Borea window is activated. Players come back to Borea after installing a new release, so that's the natural moment. Reading the file version of one DLL is cheap.
- When the version changed:
  - update `InstalledVersionText`,
  - run `RefreshCompatibilityAsync` so the Discover chips and the "Hide incompatible" filter follow,
  - refresh the open content page's compatibility chip.
- When it didn't change, do nothing, so focusing the window doesn't rebuild lists.

### Out of scope

- Updating installed mods after a KSA update. That's `borea update` and a future App update action.
- Watching the game folder with a `FileSystemWatcher`. Window activation covers the common case without holding a handle on the game directory.
- Telling the user a newer KSA release exists. `LatestVersionPing` already knows the latest build and `borea game version` shows it; showing that in the App would be its own issue.

### Done when

- A view model test: change the version the provider reports, raise the activation hook, and the chip text and a Discover row's compatibility follow.
- A second activation with an unchanged version leaves the Discover rows untouched.

averageksp:
## Notice a KSA update when the window is activated

Closes #169.

After a new KSA release was put in place while the App was open, the "KSA …" chip and the compatibility on Discover kept showing the old build until a restart.

- The window's `Activated` event calls `RefreshInstalledGameAsync`, which reads `KSA.dll` again. Players come back to Borea after installing a new release, so that is when it runs.
- When the version text changed, the chip updates and `RefreshCompatibilityAsync` evaluates every listing again, which also moves the "Hide incompatible" filter and the content page's chip.
- When it did not change, nothing is touched, so focusing the window does not rebuild lists.

Tests use the version fixture assemblies the Storage tests already build, copied as `KSA.dll` into the saved game folder: a new build updates the chip and the compatibility, an unchanged build leaves the rows alone, and a build whose version does not parse shows its text with unknown compatibility.


Maximilian-Nesslauer (changes requested):
lgtm but the App test project copies the fixture target from Borea.Storage.Tests.csproj word for word. A shared .targets file would keep the two from drifting apart.

Maximilian-Nesslauer (approved):

PR #183, opened 2026-09-13
Tell the user when a newer Borea release exists

### LINKED ISSUE #150: Tell the user when a newer Borea release exists
Borea never tells a user that a newer Borea release exists, so people keep running old builds until they look at GitHub themselves.

## Current status

#145 shows the running Borea version on the About tab, read from the informational version the release workflow stamps.
Nothing compares it with the published releases. `LatestVersionPing` in `Borea.Network/MasterServer` asks the KSA master server for the newest game build, and nothing does the same for Borea.

## What it should be instead

| Today | Should be |
|---|---|
| No release check. | A Core interface next to `ILatestVersionPing`, implemented in `Borea.Network`, reads `GET /repos/KSAModding/Borea/releases/latest` and returns the newest release version and its page URL. |
| No comparison. | The running version and the release version compare as SemVer with `ModVersion`, build metadata ignored, so `0.4.0+abc123` equals the `v0.4.0` release. |
| Nothing in the App. | When a newer release exists, the App shows a notice with the version and a link to the release page, on Home and on the About tab. It does not block anything. |
| | A failed request (offline, rate limit) shows nothing and is not an error. |
| | The check runs once per App start. GitHub allows 60 unauthenticated API requests per hour for each IP address, so it must not poll. |
| | An App preference "Check for updates at start", on by default, lets a user turn the request off. |

## Tests

- A newer release produces a notice, an equal or older release does not.
- `0.4.0+abc123` against the `v0.4.0` release produces no notice.
- A failed request produces no notice and no error.
- With the preference off, no request is made.

## Not in scope

- Downloading or installing the update. The builds are not signed yet (#76), so the user downloads from the release page.
- Release channels for Borea itself (#146), which build on this check.
- An update check in the CLI.

## Depends on

The version reading in #145.

Maximilian-Nesslauer:
Borea now asks GitHub once at start whether a newer Borea release exists. If one does, Home and the About tab show a notice with the version and a button that opens the release page. The notice blocks nothing.

Example: you run Borea 0.4.0 and 0.5.0 is published. Home then shows "A newer Borea release is available: 0.5.0" with "View release".

- One request to `GET /repos/KSAModding/Borea/releases/latest` per start, never polled. GitHub allows 60 unauthenticated requests an hour for each address.
- Versions compare as SemVer with `ModVersion`, and build metadata is ignored, so `0.4.0+abc123` equals the `v0.4.0` release. The endpoint skips drafts and pre-releases.
- No release yet (404), offline, a rate limit or a tag that is not a version: no notice and no error.
- General settings get "Check for updates at start", on by default. With it off, no request is made. The preference file keeps format version 1, and a file without the new field reads as on.

Closes #150


averageksp (approved):
Lgtm

PR #184, opened 2026-09-13
Read download counts from the index snapshot

### LINKED ISSUE #181: Read download counts from the index snapshot
RFC 0052 (content-manager-design#52) adds an optional `downloads` field to each index listing, with a `total`, per-host values and per-version entries, but Borea ignores it, so neither the CLI nor the App can show or sort by downloads.

## Current status

- `ListingParser` in `Borea.Storage/Index` does not know `downloads` and skips it.
- `ContentIndexListing` in Core has no place for counts.
- `borea search` and `borea show` have no count to print.
- The index does not publish the field yet. #126 read the curated tags against the contract before the index published them, and this issue follows that pattern.

## What it should be instead

| Today | Should be |
|---|---|
| `downloads` is ignored. | Storage reads it into an immutable Core type on `ContentIndexListing`, next to `IndexStatus`. |
| No notion of an unknown count. | An absent field, host or version entry means unknown and never zero. |
| Nothing to report. | A malformed part keeps the listing and its releases and adds a `downloads` diagnostic. A bad listing total drops the listing's counts, and a bad release entry drops only that version's count. |
| No counts in the CLI. | `borea search` shows the listing total, `borea show <id>` shows the total, each host and each listed release, and `--json` carries the values. The default order of `search` stays by id. |
| | `borea index validate` reports how many listings have counts and how many counts are malformed. |

## Tests

- Storage: a listing with full counts, without counts, with one host, a release without a version entry, a malformed `downloads` beside a valid sibling, a bad release entry beside a valid total, and a tombstone.
- Core: the constructor rejects a negative count, a total that is not the sum of its hosts, and a duplicate version.
- CLI: text and JSON output of `search` and `show` with and without counts, and the `index validate` summary.

## Not in scope

- Showing and sorting by counts in the App, which is a later task with averageksp.
- A sort option in the CLI.
- Producing the counts, which happens in content-index-releases.

## Depends on

Nothing. The field is optional, so Borea can read it before the index publishes it.

Maximilian-Nesslauer:
Borea now reads the optional `downloads` field that RFC 0052 adds to each index listing.
A count that the index does not report stays unknown and is never shown as zero.
A malformed count never makes a listing uninstallable: Borea drops only the bad part and reports a `downloads` diagnostic.

`borea search` shows the listing total, and `borea show <id>` shows the total, each host, and each listed release that has a count. `--json` carries the same values in the snapshot's shape, and `borea index validate` reports the counts.

Example, output captured from a CLI run on the test fixtures (the `show` output starts after the listing details):

```text
$ borea search flight
flight-tools  Flight Tools  2.0.0  compatible  1200 downloads

$ borea show flight-tools
Downloads: 1200 (github 479, spacedock 721)
Releases:
  2.0.0  compatible  stable
    Game: 2026.8.22.5348 to open
    Downloads: 40 (github 17, spacedock 23)
  1.0.0  compatible  stable
    Game: 2026.8.22.5348 to open
```

Release 1.0.0 has no count line because the fixture has no count for that version.

The index does not publish the field yet, and this can merge before it does.
Showing and sorting in the App is a separate task.

Closes #181


averageksp (approved):
Lgtm

PR #188, opened 2026-09-13
Post each new release to Discord

### LINKED ISSUE #149: Post each new release to Discord
When the release workflow publishes a new Borea release, nobody hears about it until someone posts a link by hand.

## Current status

`release.yml` creates the GitHub release in its last job. The only Discord integration is `notify-failure.yml`, which posts a failed `build` run through `DISCORD_WEBHOOK_URL`.

## What it should be instead

| Today | Should be |
|---|---|
| A published release is silent. | A new `announce` job in `release.yml` posts one short message to the release channel on Discord. |
| One webhook secret, for build failures. | A separate secret, `DISCORD_RELEASE_WEBHOOK_URL`, for the release channel. Without it the job skips and the release still succeeds. |
| The release body is long (checksum table, first-start steps, provenance). | The message is short: the version, a pre-release marker for a tag with a hyphen, the link to the release page, one line for the App and one for the CLI archives, and the first few "What's Changed" entries followed by "and N more". |

## Constraints

- The post has to be a job in `release.yml`. A separate workflow on the `release: published` event would never start, because the release is created with `GITHUB_TOKEN` and GitHub starts no workflow for events caused by that token.
- Post only when the run created the release. A re-run for an existing tag replaces files and notes and must not post again, and a run started by hand publishes nothing and must not post. The create step sets an output such as `created=true`, and `announce` runs only on that.
- Build the JSON with `jq -n`, as `notify-failure.yml` does, because the change list contains pull request titles.
- Stay under the Discord limits: 2000 characters for `content`, 4096 for an embed description.
- `announce` needs no token permissions (`permissions: {}`), only the secret.

## Tests

- A run started by hand does not post.
- A re-run for a tag that already has a release does not post.
- A missing secret skips the job without failing the run.
- A long change list is cut before the Discord limit.

## Not in scope

- A Discord bot, or posts for anything other than a published release.
- Changing the release body on GitHub.
- Release channels for Borea itself (#146).

## Depends on

#135, which rewrites the release workflow, so this job is added on top of it instead of conflicting with it.

Maximilian-Nesslauer:
Closes #149.

When the release workflow creates a new release, a new `announce` job posts one short message to the release channel on Discord.

It posts only when this run created the release. A re-run for a tag that already has a release, and a run started by hand, post nothing. The job has no token permissions.

The job posts through a webhook or a bot, whichever is configured:

| Mode | Secret | Variable |
|---|---|---|
| Webhook | `DISCORD_RELEASE_WEBHOOK_URL` | none |
| Bot | `DISCORD_RELEASE_BOT_TOKEN` | `DISCORD_RELEASE_CHANNEL_ID` |

With neither secret, the job prints that nothing is configured, and the release still succeeds. When both are set, the bot posts and the run shows a warning. The bot token goes to curl through stdin and is never printed. Use a small bot only for this post, because the token gives the workflow everything that the bot can do.

To notify a role, set the variable `DISCORD_RELEASE_ROLE_ID`. The message then starts with the role mention, and `allowed_mentions` allows that role and nothing else, so a pull request title cannot notify anyone. A pre-release posts without the ping. For the bot, the role can stay not mentionable when the bot has the "Mention @everyone, @here, and All Roles" permission. For the webhook, make the role mentionable. `CONTRIBUTING.md` lists the setup for each mode.

When Discord rejects the post, or a channel or role id is not a Discord id, the job fails and the release run turns red, although the release is already published. The log then shows Discord's error code and message. The webhook call waits for Discord to save the message, so a post that is not saved also fails. The post is not retried, so it cannot appear twice.

For example, with `DISCORD_RELEASE_ROLE_ID` set (output of the test fixture, not of a real release):

```text
<@&123456789012345678> **Borea 1.2.0**
https://github.com/KSAModding/Borea/releases/tag/v1.2.0
App with the `borea` command: `Borea-1.2.0-<platform>` for Windows, Linux and macOS
CLI only: `Borea-Cli-1.2.0-<platform>` for the same platforms

Changes:
- Include the CLI in the desktop App archives (#156)
- Add an About tab to the settings (#145)
```

with `"allowed_mentions": {"parse": [], "roles": ["123456789012345678"]}`.

Before the first release with this job, someone with admin rights has to add the secret and variable of one mode, and optionally the role id.


averageksp (commented):
What about a role for notifying users when a release happens? Could that be configured?

Maximilian-Nesslauer:
> What about a role for notifying users when a release happens? Could that be configured?

Yes the webhook can do it.

The only Discord setting that you would need to ensure is that the role must allow anyone to mention it.
The Discord docs say a role mention notifies its members only when the role is mentionable or the sender has the Mention Everyone permission, and a webhook has no permissions of its own.
So every member of the server can also ping that role, and it should be a role that people join on purpose.

... which would make it a good idea to do this through the Hunter bot instead, because a bot can have the "Mention @everyone, @here, and All Roles" permission, so the role can stay not mentionable and only the bot can ping it.

only issue is that i have no idea how Hunter works and no access to it at all.

Maximilian-Nesslauer:
i am going to update this PR so that both a webhook and bot will work :)

averageksp (approved):

Maximilian-Nesslauer:
@averageksp  merged, thanks for the review!

The last step needs admin rights on the repo and access to the bot, so you will have to do it.

In GitHub: Under Settings > Secrets and variables > Actions, set one of these two modes:
- Bot (recommended, so the role can stay not mentionable): secret `DISCORD_RELEASE_BOT_TOKEN`, variable `DISCORD_RELEASE_CHANNEL_ID`. The bot needs View Channel, Send Messages, and "Mention @everyone, @here, and All Roles" in the release channel. (See PR body)
- Webhook: secret `DISCORD_RELEASE_WEBHOOK_URL`. For a role ping, the role must then be mentionable.

Optional: variable `DISCORD_RELEASE_ROLE_ID` for the role ping. Pre-releases do not ping.

Note that the existing `DISCORD_WEBHOOK_URL` secret is only for the failure notifications and the release post does not use it. CONTRIBUTING.md has the full setup.

PR #190, opened 2026-09-14
Added a Launch Active Instance split button to the home page.

### LINKED ISSUE #133: Launch Vanilla KSA using the original KSA instance
# Feature
Borea currently has no way of launching KSA using the original instance in `EnvSpecialFolders.LocalDocumentsFolder/My Games/Kitten Space Agency/`. This issue is to add the ability to launch KSA using this instance without a mod loader. It should find where `KSA.exe` is and run that. No instance pathing or anything, just base KSA.

## CLI
I would recommend putting this under the `GameCommands` in the CLI as that is where KSA information is interacted with. As Borea is not affecting anything about KSA in this situation, other than launching the game. I recommend to put it here as to not confuse the vanilla launch with the modded or content added launches which `launch <instance> <mod loader>` does.

MrJeranimo:
It has a flyout button which also lets the user launch Vanilla KSA if wanted. The button uses the standard cta button style with some height changes to match the "Home" icon and text. Dutch translation was done using Google Translate (please don't hate me).
 
 
Screenshot 2026-09-13 224108
Screenshot 2026-09-13 224113

Just a note, I could not figure out how to change the color of the chevron in the split button unfortunately.


averageksp (commented):
Its German not Dutch 🤣

src/Borea.App/Views/Pages/HomePage.axaml
```diff
@@ -8,10 +8,25 @@
...
-                <Path Classes="icon" Data="{StaticResource Icon.Home}" />
-                <TextBlock Classes="label-lg" Text="{Binding Localization.PageHomeHeading}" VerticalAlignment="Center" />
-            </StackPanel>
+			<Grid ColumnDefinitions="*,Auto" Height="24">
+				<StackPanel Orientation="Horizontal" Spacing="4">
+					<Path Classes="icon" Data="{StaticResource Icon.Home}" />
+					<TextBlock Classes="label-lg" Text="{Binding Localization.PageHomeHeading}" VerticalAlignment="Center" />
+				</StackPanel>
+
+				<SplitButton Grid.Column="1" Classes="launch" Command="{Binding PlayCommand}" IsEnabled="{Binding !IsLaunching}" IsVisible="{Binding HasActiveInstance}">
```
Maximilian-Nesslauer: [should fix] `LaunchMessage` is only shown on the instance page, so on Home neither result is visible: "Borea does not know where the game is installed", a missing executable, or "Started the game without a mod loader". The message then appears later on the instance page, where it no longer fits. Could Home show `LaunchMessage` under the header too?

Maximilian-Nesslauer: [should fix] `IsVisible="{Binding HasActiveInstance}"` also hides the second option, although starting the game without a loader needs no instance.

src/Borea.App/Localization/Resources/Resources.resx
```diff
@@ -486,4 +591,10 @@
   <data name="InstallMinutesLeftFormat" xml:space="preserve">
     <value>about {0} min left</value>
   </data>
-</root>
+  <data name="GameLaunchVanilla" xml:space="preserve">
+    <value>Launch Vanilla KSA</value>
```
Maximilian-Nesslauer: [should fix] "Vanilla" promises more than this does. The game starts without a mod loader in the shared profile, but asset mods that are enabled in that profile's manifest still load, because they need no loader (RFC 0031, `[loader]`). #157 uses "without a mod loader" for the same reason. Suggestion: "Launch KSA without a mod loader".

src/Borea.App/Localization/Resources/Resources.resx
```diff
@@ -1,5 +1,110 @@
 <?xml version="1.0" encoding="utf-8"?>
 <root>
+  <!--
```
Maximilian-Nesslauer: [minor] The Visual Studio resx editor added its schema comment block to both resx files and removed the final newline. Neither file had it before. Could you drop it?

src/Borea.App/ViewModels/MainViewModel.Instance.cs
```diff
@@ -123,6 +123,16 @@ private async Task PlayAsync()
         }
     }
 
+    [RelayCommand]
+    private void PlayVanilla()
```
Maximilian-Nesslauer: [should fix] Please add a view model test, for example that without a game directory it sets the "Set the game directory" message, like the test for `Play`.

Maximilian-Nesslauer (changes requested):
Thanks, the launch button on Home is a good idea :)

Two things need a change before this merges, and added a few smaller ones inline.

1. **[must fix] The button does not launch the active instance.** `PlayCommand` launches `_selectedInstanceEntity`, and that field is only set when an instance page is opened (`OpenInstanceAsync`). On a fresh start, the Home button is visible because `HasActiveInstance` is true, but a click does nothing and shows nothing. If you opened instance B earlier while A is active, the button labelled "Launch Active Instance" starts B. The existing `Play_WithoutALoader_ShowsTheLauncherMessage` test opens the instance page first, so it does not catch this. Suggestion: give Home its own command that reads `IInstanceRepository.GetActiveInstanceIdAsync()` and loads that instance, plus a test that runs it without opening an instance page, and one with a different instance opened before.
2. **[must fix] The German resources contain Dutch text.** `Resources.de.resx` has "Vanilla KSA Lanceren" and "Actief Exemplaar Starten". The file already uses "Instanz" and "Mod-Loader", so for example "Aktive Instanz starten" and "KSA ohne Mod-Loader starten".

Maximilian-Nesslauer:
> Just a note, I could not figure out how to change the color of the chevron in the split button unfortunately.

The chevron is a PathIcon inside the SplitButton template, and it takes its color from the foreground of the secondary button. 

The SplitButton.launch style sets only Background, so the chevron keeps the default foreground of the Fluent theme.

The theme reads SplitButtonForegroundPointerOver, SplitButtonForegroundPressed and SplitButtonForegroundDisabled (plus the normal SplitButtonForeground) as dynamic resources, so overriding those keys in the SplitButton.launch style should give the chevron the same color as the play icon.

I took this from the Fluent SplitButton.xaml on the Avalonia master branch.

src/Borea.App/Localization/Resources/Resources.resx
MrJeranimo: No clue why this happened, sorry

src/Borea.App/ViewModels/MainViewModel.cs
```diff
@@ -159,11 +159,13 @@ public void SetMainWindowTasks()
     /// actions work from the Current Install card. Null when none is active.
     /// </summary>
     [ObservableProperty]
-    [NotifyPropertyChangedFor(nameof(HasActiveInstance))]
+    [NotifyPropertyChangedFor(nameof(HasActiveInstance), nameof(EnableActiveInstance))]
     private InstanceItem? _activeInstance;
 
     public bool HasActiveInstance => ActiveInstance is not null;
 
+    public bool EnableActiveInstance => ActiveInstance is not null && !IsLaunching;
```
Maximilian-Nesslauer, line 177: [minor] `EnableActiveInstance` also reads `IsLaunching`, but only `ActiveInstance` notifies it, so the button does not show as disabled while a launch runs. The command still refuses a second launch, so only the look is off. `[NotifyPropertyChangedFor(nameof(EnableActiveInstance))]` on `_isLaunching` fixes it.

src/Borea.App/Views/Pages/HomePage.axaml
```diff
@@ -8,10 +8,33 @@
...
+				<!-- Custom Split Button: Primary is the default command, Flyout is a dropdown menu -->
+				<Grid Grid.Column="1" ColumnDefinitions="Auto,Auto" >
+					<Button Grid.Column="0" Grid.Row="0" Classes="splitButton primary" Command="{Binding PlayActiveInstanceCommand}" IsEnabled="{Binding EnableActiveInstance}">
+						<StackPanel Orientation="Horizontal" Spacing="6">
+							<Path Classes="icon-filled size-16" Fill="{StaticResource Brush.SurfaceHeader}" Data="{StaticResource Icon.Play}" VerticalAlignment="Center" />
+							<TextBlock Classes="action-lg" Text="{Binding Localization.LaunchActiveInstance}" VerticalAlignment="Center" />
+						</StackPanel>
+					</Button>
+					<Button Grid.Column="1" Grid.Row="0" Classes="splitButton flyout">
+						<Path Classes="icon-filled size 16" Fill="{StaticResource Brush.SurfaceHeader}" Data="{StaticResource Icon.CaretDown}" VerticalAlignment="Center" />
```
Maximilian-Nesslauer: [minor] `size 16` is two classes, `size` and `16`, so the chevron does not get the `size-16` style. It should be `size-16`.

src/Borea.App/Localization/Resources/Resources.de.resx
```diff
@@ -486,4 +486,13 @@
   <data name="InstallMinutesLeftFormat" xml:space="preserve">
     <value>noch etwa {0} min</value>
   </data>
-</root>
+  <data name="LaunchWithoutModLoader" xml:space="preserve">
+    <value>KSA ohne Mod Loader Starten</value>
```
Maximilian-Nesslauer: [minor] German writes the verb in lowercase here, and this file uses "Mod-Loader": "KSA ohne Mod-Loader starten", and "Aktive Instanz starten" below.

src/Borea.App/Localization/Resources/Resources.de.resx
```diff
@@ -486,4 +486,13 @@
...
+  </data>
+  <data name="LaunchActiveInstance" xml:space="preserve">
+    <value>Aktive Instanz Starten</value>
+  </data>
+  <data name="LaunchWithoutModLoaderToolTip" xml:space="preserve">
+    <value>Dadurch wird KSA so gestartet, als würden Sie normalerweise die KSA.exe ausführen. Es wird keine Borea-Instanz verwendet; stattdessen läuft die gemeinsam genutzte Instanz, die von KSA verwaltet und erstellt wird.</value>
```
Maximilian-Nesslauer: [minor] The rest of this file uses "du", for example "Du siehst Inhalt aus einer Instanz". Suggestion, which also uses the word from the English comment: "Startet KSA so, als würdest du KSA.exe direkt ausführen. Es nutzt das gemeinsame Profil, das KSA selbst anlegt, und keine Borea-Instanz."

src/Borea.App/Localization/Resources/Resources.resx
```diff
@@ -486,4 +486,13 @@
...
   </data>
-</root>
+  <data name="LaunchWithoutModLoader" xml:space="preserve">
+    <value>Launch KSA without a Mod Loader</value>
+  </data>
+  <data name="LaunchActiveInstance" xml:space="preserve">
+    <value>Launch Active Instance</value>
+  </data>
+  <data name="LaunchWithoutModLoaderToolTip" xml:space="preserve">
+    <value>This will launch KSA as if you normally ran KSA.exe. It does not use a Borea instance, it runs the shared instance that KSA manages and creates.</value>
```
Maximilian-Nesslauer: [minor] Borea calls this the shared profile elsewhere (`borea game launch` in #157), because an instance is Borea's own folder. Suggestion: "Launches KSA the same way as running KSA.exe. It uses the shared profile that KSA creates, not a Borea instance."

tests/Borea.App.Tests/ViewModels/InstanceViewModelTests.cs
```diff
@@ -151,4 +151,16 @@ public async Task LanguageChange_RetitlesTheGroups()
         Assert.Equal(harness.Localization.InstanceGroupMods, viewModel.ContentGroups.Single().Title);
         Assert.Equal(harness.Localization.HomeInstanceSourceCustom, viewModel.ActiveInstance!.SourceText);
     }
+
+    [Fact]
+    public async Task PlayKSAWithoutModLoader_ShowsErrorMessageIfNoGameDirectorySet()
```
Maximilian-Nesslauer, line 156: [should fix] This covers the launch without a loader. Could there also be a test for the bug that was fixed: with an active instance and no instance page opened, `PlayActiveInstanceCommand` starts that instance, and with another instance opened before, it still starts the active one? Both resx files also still end without a final newline.

Maximilian-Nesslauer (changes requested):
LGTM but a few small points remain 👍 :)

tests/Borea.App.Tests/ViewModels/InstanceViewModelTests.cs
MrJeranimo: I couldn't figure out have to get the settings repository to use test settings as the current constructor leaves the settings null.

Maximilian-Nesslauer (approved):

PR #212, opened 2026-09-14
Let a selection in the game log span lines

### LINKED ISSUE #211: Instance log: select more than one line
## Instance log: select more than one line

On an instance page's Log tab, you can only select text within a single line. Dragging across several lines stops at the end of the line you started in, so copying a stack trace or a block of warnings means copying line by line, or using "Copy" for the whole log.

### Cause

`InstancePage.axaml` renders `GameLogLines` with an `ItemsControl`, and each `GameLogLine` gets its own `SelectableTextBlock` (with coloured runs for the time, level, tag and message). A selection in Avalonia belongs to one `SelectableTextBlock`, so it can't continue into the next item.

### Proposal

Render the log as **one** `SelectableTextBlock` whose inlines hold every line:

- For each `GameLogLine`, add the same runs as today (`log-muted` time, level coloured by severity, `log-tag`, message), followed by a `LineBreak`.
- Build the inlines when `GameLogLines` changes (load, reload), in the page's code-behind or in a small attached behaviour, since `Inlines` can't be bound to a collection in XAML.
- Keep the `log-*` classes, so the colours and the theme stay as they are.
- Copying a selection then copies exactly the selected text across lines, with line breaks.

A read-only `TextBox` would also allow multi-line selection, but it loses the per-level colours, so the single text block is the better fit.

### Watch out for

- Very long logs: one text block with thousands of runs has to lay out in one go. If the reload feels slow on a real KSA log, cap the view to the last N lines (for example 5000) with a note "Showing the last 5000 lines", and keep "Copy" and "Open" working on the whole file.

### Done when

- On the Log tab, a drag across several lines selects all of them, and Ctrl+C copies them with their line breaks.
- Levels keep their colours in both themes.
- Reload replaces the text without leaving a stale selection.

averageksp:
## Let a selection in the game log span lines

Closes #211 

The Log tab rendered each line as its own `SelectableTextBlock` inside an `ItemsControl`, and a selection cannot cross from one control into the next. Selecting a stack trace stopped at the end of the first line.

- New `GameLogView`, a `SelectableTextBlock` that builds its inlines from `GameLogLines`: time, level, tag and message as runs with the same `log-*` classes, and a `LineBreak` between lines. It uses the `SelectableTextBlock` style key, so the theme and the `log-line` style still apply.
- A load clears the collection and adds the lines one by one; the view rebuilds its text once afterwards on the dispatcher instead of once per line. A rebuild drops the old selection.
- `InstancePage` uses it in place of the `ItemsControl`. The view model is unchanged.

The reader already caps the tab at 200 lines, so one text block stays cheap to lay out.


Maximilian-Nesslauer (dismissed):
LGTM 👍 

Two minor things: GameLogView could unsubscribe from CollectionChanged when it is detached from the visual tree, and the class comment should describe only the current behaviour ("one selectable text, so a selection can span lines") without the part about what stopped before.

Maximilian-Nesslauer (approved):

PR #214, opened 2026-09-14
Open the mod page from an instance

### LINKED ISSUE #191: Open the mod page from an instance
### what

The instance page lists the installed mods, but a click on a mod does nothing.
A click should open the same mod page that Discover opens.

| Today | Should be |
|---|---|
| The icon and the name of a row on the instance page have no action, because `ContentItem` has no open command and `OpenContentAsync` accepts only a `DiscoverItem`. | A click on the icon or the name opens the mod page. |
| The breadcrumb of the mod page always says "Discover". | When the page was opened from an instance, the breadcrumb leads back to that instance. |
| | A mod that is not in the content index, or a mod that Borea did not install, has no link, and a tooltip says why. |

### Tests

- A click on an index mod opens its mod page, and the breadcrumb returns to the instance.
- A mod without a listing opens nothing.

### Not in scope

- A page for a mod that is not in the content index.

averageksp:
## Open the mod page from an instance

Closes #191.

Clicking a mod on the instance page did nothing. Now:

- The icon and the name of a row open the same mod page Discover opens, for mods Borea installed that are in the content index. The instance page loads the Discover listings for this, through the same shared load Discover uses.
- Opened from an instance, the page matches `library-instance-content` in #8: the breadcrumb reads "Library / <instance> / <type>", a banner under it says "You're viewing content inside an instance." with a "View instance" link back, and the Library rail item stays lit. Opened from Discover, the page is unchanged.
- Rows without a page (not installed by Borea, or not in the index) show no link and say why in a tooltip.
- Instance rows now refresh their author text and the tooltip when the language changes.

Two small design fixes on the way:

- The separator on the nav rail uses the design's grey (`Brush.BorderStrong`) instead of black.
- Link buttons center their content vertically, so a link next to text lines up with it.

English and German strings. Tests: an indexed Borea mod opens its page and returns to the instance, a mod not installed by Borea has no link, a later open from Discover keeps the Discover breadcrumb, and the tooltip follows a language change.


src/Borea.App/ViewModels/MainViewModel.Content.cs
```diff
@@ -79,6 +93,29 @@ internal async Task OpenContentAsync(DiscoverItem item)
...
+    /// <summary>
+    /// Opens the page of an installed mod from the instance page. The
+    /// breadcrumb then names the instance and leads back to it.
+    /// </summary>
+    internal async Task OpenContentFromInstanceAsync(DiscoverItem item)
+    {
+        if (item is null || SelectedInstance is null)
+            return;
+
+        ContentReturnInstance = SelectedInstance;
```
Maximilian-Nesslauer, line 125: [should fix] Add, the Installed state and Remove on this page still act on `ActiveInstance`, while the breadcrumb and the banner name this instance. When the instance is not the active one, the page installs into or removes from a different instance than the one it names. Either act on this instance, or hide Add and Remove on a page opened from an instance that is not active. A test that opens the page from a second, inactive instance would pin it.

src/Borea.App/ViewModels/MainViewModel.Content.cs
```diff
@@ -79,6 +93,29 @@ internal async Task OpenContentAsync(DiscoverItem item)
...
+        await ShowContentAsync(item);
+    }
+
+    [RelayCommand]
+    private Task ReturnToInstanceAsync() =>
```
Maximilian-Nesslauer: [should fix] `ContentReturnInstance` is the `InstanceItem` from when the page opened. After an install or remove on the page, `ReloadInstancesAsync` rebuilds `Instances`, so this reopens the old object and the header shows the old mod count (`ModCount` is set in the constructor). Looking the instance up again by id fixes it, for example `Instances.FirstOrDefault(instance => instance.InstanceId == ContentReturnInstance.InstanceId)`, or keep only the id.

Maximilian-Nesslauer (changes requested):
Thanks for the PR :)

I added 2 things that should be fixed inline.

Maximilian-Nesslauer (approved):

PR #219, opened 2026-09-14
Read listing images and dates from the index snapshot

### LINKED ISSUE #197: Read listing images and dates from the index snapshot
### what

RFC 0058 adds an optional icon, description images, and the two dates `published_at` and `updated_at` to the content index.
Borea ignores all of them.
This issue only reads them into Core, so that the App and the CLI can use them later.
Like #181, Borea can read the fields before the index publishes them.

| Today | Should be |
|---|---|
| `[images]` in an authored document is skipped. | Storage reads the icon and the description image records into immutable Core types, with `url`, `sha256`, `width`, `height`, `size`, and the optional `license`, `attribution` and `source`. |
| The derived dates are skipped. | A listing and a pack entry carry `PublishedAt` and `UpdatedAt`, which are absent when the snapshot has none. |
| | A malformed image record or date keeps the listing usable, drops only the bad part, and adds a diagnostic. |
| `borea show <id>` prints no dates. | It prints both dates when they are present. |

### Tests

- A listing with an icon and description images, and one without.
- A malformed record next to a valid one.
- Both dates, no dates, and only `published_at`.
- A tombstone.

### Not in scope

- Downloading or showing images, which is a later App issue.

### Depends on

Nothing. Related to KSAModding/content-manager-design#62.

Maximilian-Nesslauer:
Closes #197

Borea now reads the icon, the description images and the `published_at` and `updated_at` dates from the index snapshot into Core.
Images attach to a listing and to each pack version, and dates attach to a listing and a pack.
A bad image record or date drops only itself and adds a diagnostic with the new `images` or `dates` scope, so the listing stays usable.
`borea show <id>` prints `Published:` and `Updated:` when the index has them, and `borea index validate` counts malformed images and dates.
Nothing downloads or shows images yet.


averageksp:
This is what ive been waiting for 🔥

averageksp (approved):

PR #220, opened 2026-09-14
Tell the player when the loader stops right after Play

### LINKED ISSUE #216: Tell the player when the loader exits right after Play
## Tell the player when the loader exits right after Play

When a mod breaks StarMap on start, pressing Play looks like nothing happened. Borea says "Started StarMap for instance 'Main'", a console window may flash, and that's it. The game never opens, and the player has no idea why or what to do.

### Reproduced

- KSA 2026.9.7.5402, StarMap 0.4.6, KSArmory 0.8.44 installed in the instance.
- `borea launch "Main KSA Instance" StarMap` prints `Started StarMap …` and exits 0.
- StarMap is gone a moment later with exit code `-532462766` (`0xE0434352`, an unhandled .NET exception). The instance's `logs/KittenSpaceAgency.log` isn't touched, because the game never started.
- Running `StarMap.exe -InstancePath <instance>` by hand shows the reason on stderr:

```
Unhandled exception. System.Reflection.ReflectionTypeLoadException: Unable to load one or more of the requested types.
Method 'DrawAxes' in type 'KSArmory.RoundFollowable' from assembly 'KSArmory, Version=0.8.44.0, Culture=neutral, PublicKeyToken=null' does not have an implementation.
   at StarMap.Core.ModRepository.RuntimeMod.TryCreateMod(...)
   at StarMap.Core.ModRepository.ModLoader.PrepareMods()
   at StarMap.Program.Main(String[] args)
```

KSArmory 0.8.44 was built against an older game (`game_min = 2026.8.19.5261`, no `game_max`), so Borea rates it *Compatible* with 2026.9.7, but its assembly no longer loads.

### Why Borea says nothing

- `ProcessStarter` redirects the loader's stdin, stdout and stderr, then closes Borea's ends right away, so the crash text is thrown away.
- `LoaderLauncher.Launch` returns `Success` as soon as the process exists, and `PlayAsync` shows that message. Nobody looks at the process again until the next launch (`Forget(exited: true)`).

### Proposal

**Watch the first moments of a launch.**
- Keep reading the loader's stdout and stderr instead of closing them, into a bounded buffer (the last ~200 lines) and into `<instance>/logs/borea-launch.log`, so Borea never blocks the loader on a full pipe.
- Report the launch as *starting* and watch the process for a short window (for example 20 seconds, or until the game's own log is written, whichever comes first).
- If the process exits in that window with a non-zero code, the launch result becomes a failure with the exit code and the captured output.

**Say what went wrong, in the player's words.**
- When the output has a .NET exception that names an assembly of an installed mod (`from assembly 'KSArmory, Version=0.8.44.0'`), the message names the mod: "KSArmory 0.8.44 stopped StarMap from starting. It may not work with KSA 2026.9.7. Disable it and try again, or check for an update."
- Offer actions next to the message on the instance page (and on Home once #173 lands): **Disable KSArmory**, **Show details** (the captured output, selectable and copyable), **Open launch log**.
- When no mod can be named, show "StarMap closed right after starting (exit code …)" with **Show details**.

**CLI.** `borea launch` waits the same window. On an early exit it prints the captured stderr and the named mod, and exits `1` instead of `0`.

### Related

- The compatibility rating can't catch a mod that declares no `game_max` and breaks on a newer build. Showing which mods failed to load is the fallback that works regardless of what the listing declares.
- This is also a report StarMap could handle better itself (skip the broken mod instead of crashing). Worth raising with the StarMap authors, but Borea should explain the failure either way.

### Done when

- With a fake process that writes the exception above to stderr and exits with `0xE0434352` within the window, `Launch` returns a failure naming KSArmory. The instance page shows the message with Disable, Show details and Open launch log.
- A loader that keeps running past the window is still reported as started, and its output keeps flowing into the launch log without blocking it.
- `borea launch` exits 1 and prints the reason in the same case.

averageksp:
## Tell the player when the loader stops right after Play

Closes #216.

When a mod broke StarMap on start, Play said "Started StarMap" and nothing else happened. Borea closed the loader's output right away and counted a launch as started as soon as the process existed, so the error StarMap printed was lost.

Reproduced with KSA 2026.9.7.5402, StarMap 0.4.6 and KSArmory 0.8.44: StarMap exits with 0xE0434352 and `Method 'DrawAxes' in type 'KSArmory.RoundFollowable' from assembly 'KSArmory, Version=0.8.44.0' does not have an implementation.`

### Changes

- `ProcessStarter` keeps reading the started process's output and error into a buffer of the last 200 lines instead of closing the pipes. `IStartedProcess` gains `ExitCode`, `RecentOutput` and `WaitForExitAsync(timeout)`. Input is still closed, and handle inheritance is unchanged.
- `ILauncher.WatchStartAsync(instance, started)` watches a started launch until the game writes its log, the loader exits, or 20 seconds pass. A non-zero exit in that time returns the new `LaunchOutcome.ExitedEarly` with the exit code, the output and `BlamedModId`. An exit with 0 or a loader still running stays `Started`. `Launch` itself is unchanged.
- `LoaderCrashReport` reads the assembly names from .NET exception messages. The launcher blames the first installed mod whose id matches an assembly name, or whose folder holds that DLL.
- What the loader wrote goes to `<instance>/logs/borea-launch.log` (new `IGamePathProvider.GetInstanceLaunchLogPath`), replaced on each launch. `LoggingLauncher` writes early exits to Borea's log.
- App: Play shows "Starting StarMap…" while it watches. On an early exit the instance page shows which mod stopped StarMap, with Disable <mod>, Show details and Open launch log. Disabling reloads the page and says to press Play again. English and German.
- CLI: `borea launch` waits for the watch, and on an early exit prints the reason and the loader's output on stderr and exits 1.

### Tests

- Launcher: a crash naming an installed mod, a DLL inside a mod folder with another id, a crash naming no mod, exit code 0, a loader still running after the window, the game log ending the watch early, and a failed launch passed through unchanged.
- Process starter: a real script's output, error and exit code are kept, and the wait times out on a running process.
- CLI: an immediate crash exits 1 with the output.
- Crash report parsing.

### Note

A loader that exits with 0 to restart itself counts as started. If the restarted process then crashes, Borea doesn't see it, since it only watches the process it started.


src/Borea.Storage/Launch/LoaderLauncher.cs
```diff
@@ -114,6 +133,124 @@ public LaunchResult Launch(Instance instance, ModMetadata? loader)
...
+        {
+            // the handle was released by another call while this one watched
+            return started;
+        }
+
+        var output = process.RecentOutput;
+        var exitCode = exited ? process.ExitCode : null;
+        WriteLaunchLog(_pathProvider.GetInstanceLaunchLogPath(instance.InstanceId), started.Plan, output, exitCode);
+
+        if (exitCode is null or 0)
```
Maximilian-Nesslauer: [should fix] StarMap exits with 0 when it cannot start at all: `Program.Main` returns after `LoaderConfig.TryLoadConfig` prints that `GameLocation` is empty or that `KSA.dll` is missing, and after `TryLoadCoreAndGame` prints "Game directory not found". Those cases come back as Started. A restart also exits with 0, but then the new process writes the game log, so one way is to count an exit with 0 as started only when the game log was written in the window, and otherwise return `ExitedEarly` without a blamed mod.

src/Borea.App/ViewModels/MainViewModel.Instance.cs
```diff
@@ -160,6 +192,55 @@ private async Task PlayAsync()
         }
     }
 
+    private void ShowLaunchFailure(LaunchResult result, Instance instance, ModMetadata? loader)
```
Maximilian-Nesslauer, line 201: [should fix] Could this flow get a view model test? For example: Play with a fake process that exits early naming an installed mod shows the message, `CanDisableBlamedMod` is true, and `DisableBlamedModCommand` disables the mod and shows "press Play to try again". #216 lists it as a done condition.

src/Borea.Storage/Launch/LoaderLauncher.cs
```diff
@@ -114,6 +133,124 @@ public LaunchResult Launch(Instance instance, ModMetadata? loader)
...
+        }
+        catch (Exception exception) when (exception is ObjectDisposedException or InvalidOperationException)
+        {
+            // the handle was released by another call while this one watched
+            return started;
+        }
+
+        var output = process.RecentOutput;
+        var exitCode = exited ? process.ExitCode : null;
+        WriteLaunchLog(_pathProvider.GetInstanceLaunchLogPath(instance.InstanceId), started.Plan, output, exitCode);
```
Maximilian-Nesslauer, line 219: [minor] This writes the launch log once, when the watch ends. What the loader writes afterwards stays in the buffer only, while #216 describes the output as flowing into the log. Appending lines as they arrive would match the issue, or the issue text could say that the log covers the watched start.

src/Borea.Core/Launch/LaunchOutcome.cs
```diff
@@ -25,4 +25,7 @@ public enum LaunchOutcome
 
     /// <summary>The operating system refused to start the process.</summary>
     StartFailed = 7,
+
+    /// <summary>The loader started, then stopped with an error before the game came up.</summary>
+    ExitedEarly = 8,
```
Maximilian-Nesslauer: For the rebase: `main` uses 8 for `DotnetMissing` since #225, so this becomes 9.

Maximilian-Nesslauer (changes requested):
thanks, this is useful :)

Maximilian-Nesslauer (changes requested):
Thanks, all three should-fix points are fixed :)

This conflicts with `main` now, in `BoreaServices.cs` and `ViewModelHarness.cs`, because #228 changed the same test overloads. Could you rebase? I will approve after that, because a push dismisses an approval.

For the rebase: #228 added an internal `BuildAsync` overload with `IInstallCandidateSource installCandidates`. Putting `processStarter` next to it, before the `CancellationToken`, keeps the token last like in the other overloads.

the #190 Home button repeats the older launch code without `WatchStartAsync`, so since you now merge second you should add the watch there.

Maximilian-Nesslauer (approved):

PR #225, opened 2026-09-14
Launch StarMap through dotnet on Linux and macOS

### LINKED ISSUE #167: Launch StarMap on Linux with `dotnet StarMap.dll`
## Launch StarMap on Linux with `dotnet StarMap.dll`

On Linux, Borea cannot start an instance through StarMap. The StarMap listing names `launch = "StarMap.exe"` ([provides] of RFC 0035), and `LaunchPlan.ForLoader` runs that path as the executable. On Linux StarMap has to be started as `dotnet StarMap.dll` from the loader directory, so the launch fails there.

### What works today

- Windows: `LoaderLauncher` runs `<loader dir>/StarMap.exe` with the instance handover, started in the loader directory.
- Linux: the same plan, which points at a Windows executable.
- `borea game launch` (#157) has the same gap on its side: `GameExecutable.FileName` knows only `KSA.exe`, so Linux gets "Borea does not know the name of the game's executable".

### What Linux needs

- Executable `dotnet`, arguments `StarMap.dll` followed by the handover (`--flag <instance root>` or the environment variable, exactly as now), working directory the loader directory.
- A clear message when `dotnet` is not on the `PATH`, because StarMap is framework-dependent and Borea's own self-contained runtime does not help it.

### How to get there

RFC 0035 has one `launch` target per loader and no notion of platform or runtime, so this needs a decision in content-manager-design:

1. **Per-platform targets** in `[provides]`, for example `launch.windows = "StarMap.exe"` and `launch.linux = "StarMap.dll"` with `runtime.linux = "dotnet"`.
2. **A runtime key**, `launch = "StarMap.dll"` with `runtime = "dotnet"`, which works on every platform including Windows.

Option 2 is smaller and matches how StarMap actually runs everywhere, but it changes what Windows starts, so the StarMap authors should agree.

Until the RFC settles it, Borea could special-case it without guessing much: on a non-Windows system, when `launch` ends in `.exe` and a `.dll` with the same name sits next to it, plan `dotnet <name>.dll`. That keeps Linux users unblocked and is easy to delete once the listing says it explicitly.

### Done when

- An instance with StarMap launches on Linux from the App and from `borea launch`.
- `LaunchPlan` tests cover the Linux plan and the missing `dotnet` case.
- The design decision is linked here, or the special case is documented as temporary.

Maximilian-Nesslauer:
Closes #167

On Linux and macOS, Borea now starts StarMap as `dotnet StarMap.dll` when the listing names `StarMap.exe` and `StarMap.dll` is next to it.
The handover arguments, the environment variable and the working directory stay the same, and the App and `borea launch` both use this start.
When dotnet is not on the PATH, the launch stops with a message that says so.
This is a temporary special case, and it goes once the StarMap listing says how the loader starts on each system.
Adopting a StarMap folder on Linux also reads the version from `StarMap.dll`, because the Windows app host has no version there.


averageksp (approved):
Builds are passing now, nice lgtm

PR #227, opened 2026-09-14
App one button remove

### LINKED ISSUE #192: Remove a mod with one button, the way Add works
### what

Installing a mod takes one click on "Add".
Removing it needs the three-dot menu and then "Remove".
Removing should be as easy to find as adding.

| Today | Should be |
|---|---|
| On the mod page, an installed mod shows an "Installed" chip and the three-dot menu. | The slot of "Add" shows a "Remove" button. |
| On Discover and on the instance page, Remove is only in the three-dot menu. | The row shows a remove icon button next to the "Installed" chip. |
| Remove asks for a confirmation in the row. | It still does. |
| | When Borea may not remove the mod, because another mod needs it or Borea did not install it, the button is disabled and its tooltip names the reason. |

### Tests

- Remove from the mod page works without the menu and still asks for a confirmation.
- A required dependency and a foreign mod show a disabled button with the reason.

### Not in scope

- Removing several mods at once.

averageksp:
## Remove a mod with one button, the way Add works

Closes #192. Builds on #214.

Installing took one click on Add, removing needed the "⋮" menu and then Remove. Now:

- Mod page: an installed mod shows a Remove button where Add was. The header "⋮" is gone, since Remove was its only entry.
- Discover rows: a trash button next to the Installed chip replaces the "⋮".
- Instance page rows: a trash button next to the enable toggle replaces the "⋮".
- Removing still asks for confirmation inline.
- When a mod can't be removed, the button is disabled and its tooltip says why: files Borea did not install, or another mod that requires it. The tooltip shows on the disabled button too.

`RemoveBlockedReason` applies the same ownership and `CheckUninstall` rules the removal itself uses, so every button knows ahead of time. The Discover rows use the active instance as last read, and the reasons follow a language change. Adds a Phosphor trash icon.

Tests: removing from the mod page asks first and works, a mod Borea did not install is disabled with its reason on Discover and the instance page, a mod another mod needs is disabled and names it, and the reason follows a language change.


src/Borea.App/Views/Pages/ContentPage.axaml
```diff
@@ -53,19 +75,21 @@
...
                                     <ProgressBar Width="96" Height="6" VerticalAlignment="Center" IsVisible="{Binding SelectedContent.IsRemoving}" IsIndeterminate="True" Foreground="{StaticResource Brush.Accent}" />
-                                    <Border Classes="chip positive" VerticalAlignment="Center" IsVisible="{Binding SelectedContent.IsInstalled}" ToolTip.Tip="{Binding InstalledInText}">
-                                        <TextBlock Text="{Binding Localization.DiscoverInstalled}" />
-                                    </Border>
+                                    <!-- an installed mod: Remove where Add was, disabled with the reason when it has to stay -->
+                                    <Button Classes="text danger" Height="38" Padding="16,0,20,0" Command="{Binding SelectedContent.BeginRemoveCommand}" IsEnabled="{Binding SelectedContent.CanRemove}" ToolTip.Tip="{Binding SelectedContent.RemoveToolTip}" ToolTip.ShowOnDisabled="True">
```
Maximilian-Nesslauer: [should fix] This replaces the "Installed" chip, whose tooltip named the instance ("Installed in Main"). Now nothing on the page says which instance Remove and Add act on, which matters more once #214 opens this page from an instance. Could the chip stay to the left of the button?

src/Borea.App/Views/Templates.axaml
```diff
@@ -155,13 +158,8 @@
...
                                 </Border>
-                                <Button Classes="icon" IsVisible="{Binding !IsRemoving}" AutomationProperties.Name="{Binding Name}">
-                                    <Button.Flyout>
-                                        <MenuFlyout Placement="BottomEdgeAlignedRight">
-                                            <MenuItem Header="{Binding $parent[Window].((vm:MainViewModel)DataContext).Localization.ContentRemove}" Command="{Binding BeginRemoveCommand}" />
-                                        </MenuFlyout>
-                                    </Button.Flyout>
-                                    <Path Classes="icon-filled size-18" Data="{StaticResource Icon.More}" />
+                                <Button Classes="icon" Command="{Binding BeginRemoveCommand}" IsVisible="{Binding !IsRemoving}" IsEnabled="{Binding CanRemove}" ToolTip.Tip="{Binding RemoveToolTip}" ToolTip.ShowOnDisabled="True" AutomationProperties.Name="{Binding $parent[Window].((vm:MainViewModel)DataContext).Localization.ContentRemove}">
```
Maximilian-Nesslauer: [should fix] When this button is disabled, it gets no style of its own: `Button.icon` in `App.axaml` has no `:disabled` style, so Fluent's disabled template setter paints its grey background on the `ContentPresenter`, and the trash `Path` keeps its normal stroke. Something like `Button.icon:disabled /template/ ContentPresenter` with a transparent background and a lower opacity, and the same for `Button.text.danger` on the mod page, the way `Button.tab:disabled` does it.

src/Borea.App/Views/Templates.axaml
```diff
@@ -109,13 +117,8 @@
...
                             <ToggleSwitch Classes="switch" IsChecked="{Binding IsEnabled}" Command="{Binding ToggleEnabledCommand}" VerticalAlignment="Center" AutomationProperties.Name="{Binding Name}" />
-                            <Button Classes="icon" AutomationProperties.Name="{Binding Name}">
-                                <Button.Flyout>
-                                    <MenuFlyout Placement="BottomEdgeAlignedRight">
-                                        <MenuItem Header="{Binding $parent[Window].((vm:MainViewModel)DataContext).Localization.ContentRemove}" Command="{Binding BeginRemoveCommand}" />
-                                    </MenuFlyout>
-                                </Button.Flyout>
-                                <Path Classes="icon-filled size-18" Data="{StaticResource Icon.More}" />
+                            <Button Classes="icon" Command="{Binding BeginRemoveCommand}" IsEnabled="{Binding CanRemove}" ToolTip.Tip="{Binding RemoveToolTip}" ToolTip.ShowOnDisabled="True" AutomationProperties.Name="{Binding $parent[Window].((vm:MainViewModel)DataContext).Localization.ContentRemove}">
```
Maximilian-Nesslauer: [minor] `AutomationProperties.Name` is "Remove" on this button and on the Discover one, so a screen reader cannot tell the rows apart. The mod name as the name and `RemoveToolTip` as `AutomationProperties.HelpText` would keep both.

Maximilian-Nesslauer (changes requested):
Thanks :)

I found Three smaller things, details inline.

Maximilian-Nesslauer (approved):
The three review points are fixed, and main is merged in. Merging on behalf of averageksp, as he asked.

PR #229, opened 2026-09-14
Update installed mods from the instance page

### LINKED ISSUE #178: App: update installed mods
## App: update installed mods

The CLI keeps mods current with `borea update [mod-id]`. The App has nothing for it: no sign that a newer release exists, no update button, no "Update all". A player who doesn't use the terminal stays on the release they first installed.

### What exists

- `borea update` builds `RequestedMod(installed.Metadata, reason, Exact: false)` for each Borea-owned mod and lets the planner pick the newest release that fits, dependencies included. `InstallPlanExecutor` replaces each mod through `IModReplacer`, which keeps a recovery folder until the swap succeeds.
- The App already plans and executes installs the same way (`PlanInstallAsync`), with warnings to confirm and errors on the row, and progress since #148.

### Proposal

- **Update available.** When an instance page opens (and after the index refreshes), compare each Borea-owned mod with the newest release the planner would pick. Rows with a newer release get an update icon (⇄ in the design) with the new version in the tooltip.
- **Per mod.** The icon plans an update of that mod and runs it. Warnings wait for "Update anyway", like installs do.
- **Update all.** A button on the instance page header when at least one update exists. It plans every owned mod together, so shared dependencies are resolved once.
- **Home.** The Current Install card shows "3 updates" next to the mod count, linking to the instance page.
- Foreign mods are not updated and show no icon.

### Out of scope

- Updating on a schedule or at startup without asking.
- Release channels (#151). Updates follow whatever "newest" means once that lands.

### Done when

- View model tests: a newer release shows the icon, updating one mod replaces it and clears the icon, "Update all" plans all owned mods in one plan, a planner conflict shows on the row and changes nothing.

Maximilian-Nesslauer:
Closes #178

When an instance page opens, Borea checks each mod it installed for a newer release in the saved release channel, and the row shows an update icon.
The icon updates that mod, and "Update all" in the header plans all these mods together, so a shared dependency is resolved once.
As in the design in #8, the icon sits after the version in the slot of the dependency lock, and a dependency row has no icon of its own, because "Update all" updates it.
"Update all" is in the page header as #178 asks, while the design puts it into the instance menu.
Warnings wait for "Update anyway" and errors show on the row, like installs on Discover, and the other actions of the page wait while an update runs.
The icon also shows when the newer release has a conflict, and the click then shows that conflict.
The "N updates" chip on the Home card is not in this pull request, because #190 changes that card.


averageksp (dismissed):

averageksp (dismissed):

src/Borea.App/Localization/Resources/Resources.de.resx
MrJeranimo, line 1: Imma trust the German guy

averageksp (dismissed):

MrJeranimo (approved):
LGTM, worked on my machine too.

PR #287, opened 2026-09-17
Keep the line of rounded thumbnails, tables and panels whole

### LINKED ISSUE #270: Black corners on rounded containers
## Black corners on rounded containers

Every rounded container that clips its content shows black corners where the page background should be. It's visible on the mod page (the 170 px icon and the version table), on the 48 px thumbnails in the instance and Discover rows, and anywhere else a `Border` with a `CornerRadius` has `ClipToBounds="True"` and content that paints its own background. There are 15 of those in `src/Borea.App/Views`.

### What's going on

`ClipToBounds` clips to the rectangle, not to the rounded shape. The child's own background (the image view's `Background`, the table's header row) fills the square, and the rounded border is drawn on top. What shows outside the curve is whatever the child painted there, and the corner comes out darker than the page instead of blending into it.

### Fix

Two parts, both small:

1. In XAML, stop relying on `ClipToBounds` for rounding. Give the outer `Border` the background and let the child be transparent, or set `Clip` on the child to a `RectangleGeometry` with the same radius. A quick pass over the 15 places, checking each in Borealis and Light.
2. `ListingImageView` gets a `CornerRadius` property and pushes a rounded clip in `Render` before it fills the background and draws the bitmap. Then the thumbnails don't need `ClipToBounds` at all.

A helper style like `Border.rounded-clip` could carry the clip so the fix isn't repeated per view.

### Check

- Mod page header, instance rows, Discover rows and the version table in both themes.
- Icons with transparent corners (the center crop from #266) show the thumbnail background in the corners, not black.
- Zoom at 125% and 150% to make sure the clip edge is antialiased.

written by an LLM

Maximilian-Nesslauer:
Closes #270

**For players**

Rounded thumbnails, tables, modals and side panels lost their border line in the corners.
The content inside painted over the line there, so the corners looked dark and cut off.
Now the line runs unbroken around every corner, in Borealis and in Light.
This affects the 170 px icon on the mod page and the pack page, the 48 px icons on the instance page, the 100 px icons on Discover and in pack rows, and the images in a description.
It also affects the versions table, the instance content and game save tables, the three pack tables, the modals and the side panels.
An icon with transparent corners shows the slot background in its corners.
The clip edge stays antialiased at 125 and 150 percent display scaling.

**The cause**

The issue says that `ClipToBounds` clips to a rectangle.
In Avalonia 12 that is not the case.
On a `Border`, `ClipToBounds` clips to the rounded outer edge (`CompositionBorderVisual.ServerBorderVisual.PushClipToBounds`).
But the child sits inside the line with square corners, and it draws after the border.
So in each corner, the background of the child covers the line up to the outer edge.

What reproduced here is the broken line, not black pixels, neither in headless Skia nor in an offscreen ANGLE render on Windows.
@averageksp, could you check on your machine whether the black corners are gone too? If you still see black outside the curve, the cause is elsewhere and this issue should stay open.

**For contributors**

`ListingImageView` has a new `CornerRadius` property.
`Render` pushes a rounded clip before it fills the background and draws the bitmap.
The new style `Border.thumbnail > views|ListingImageView` sets the radius to 7.5, which is the thumbnail radius 8 less half of its 1 px line.
Thumbnails no longer set `ClipToBounds`.

The new attached property `RoundedClip.IsEnabled` on a `Border` sets the `Clip` of its child to the inner edge of the line.
The clip radius is the corner radius less half the line thickness, with the same layout rounding that `Border` uses.
The clip follows changes to the size, the line thickness, the corner radius and the child.
Use it instead of `ClipToBounds` on a rounded `Border` that has a line and a child that paints its own background.
`Border.modal`, `Border.side-panel` and the six tables use it.
`Border.card` keeps `ClipToBounds`, because it has no line.

`Borea.App.Tests` now references `Avalonia.Headless`.
The new helper `HeadlessApp` starts a headless Skia session with the App styles, so a test can render a view and read its pixels.
These render tests run on the Linux and macOS runners for the first time with this pull request.

An LLM was used for this


averageksp (approved):
Yup, now its fixed. Thank you Maxi :)

PR #288, opened 2026-09-17
Fade listing images in from a skeleton

### LINKED ISSUE #286: Fade description images in from a skeleton
### what

Description images appear with a hard flash when they finish loading (beik on Discord, 2026-09-17).

| Today | Should be |
|---|---|
| The slot shows the placeholder and then switches to the image at once. | While the image loads, the slot shows a skeleton with a soft pulse in the size of the record. When the image is ready, it fades in over a short time. |
| | A failed image keeps the missing-image placeholder with its alternative text, with no pulse. |
| | Icons in lists use the same fade, so rows do not flash while scrolling. |
| | The pulse and the fade stop when the setting to load no images is on, and they respect the system setting for reduced motion where Avalonia exposes it. |

### Tests

- The view switches from skeleton to image and from skeleton to the failure placeholder.

Depends on #270, which changes the same image view.

Maximilian-Nesslauer:
Closes #286

This builds on #287, which is merged, so it targets `main`.

### What changes for players

Icons and description images no longer appear with a hard flash.
While an image loads, its slot shows a skeleton in the size of the image.
When the load takes longer than 0.3 seconds, the skeleton pulses softly.
When the image is ready, it fades in over 0.2 seconds.
An image that loads fast, for example from the cache, fades in over 0.1 seconds and does not pulse.
An image that fails shows the missing-image placeholder, with no pulse.
With "Load images from author hosts" off, the slot shows the placeholder at once, and a cached image shows with no fade.
This applies to list rows, page headers, Home tiles and description images.
The rounded corners stay on the skeleton, the image and the placeholder.

### What changes for contributors

`ListingImageView` has three states: placeholder, loading and loaded.
The skeleton is a panel inside the view that uses the view `Background`, with a pulse layer in the new `PulseBrush`.
`App.axaml` sets `PulseBrush` to `Brush.SurfaceRaised`, so both themes take the pulse color from their tokens.
The pulse is a composition animation on the render thread, not a timer per view.
It runs only while the view is loading, attached and near the viewport.
It also ends by itself after about 30 seconds, because a hidden page and a load that never reports back do not stop it.
The fade is an opacity transition that hides the skeleton over the image.
The rounded clip moved from `Render` to the `Clip` of the view, so it covers the skeleton and the placeholder too.
`ListingImage.LoadsFromAuthorHosts` tells the view if the image setting is on.
Avalonia 12.1.2 does not expose the system setting for reduced motion, so the pulse and the fade are short and soft instead.


averageksp (approved):
thanks 👍

averageksp (dismissed):

averageksp (approved):

PR #313, opened 2026-09-18
Choose the folder where Borea keeps instances and backups

### LINKED ISSUE #273: Choose the folder where Borea keeps instances and backups
### what

Borea keeps every instance and every backup in `%LocalAppData%\Borea` on the system drive.
An instance with a large mod quickly takes gigabytes (KSP-Redux 1.3.2 is 1.86 GB), and many players have more free space on another drive.

| Today | Should be |
|---|---|
| `Instances` and `Backups` are always in `%LocalAppData%\Borea`. | A "Library folder" setting decides where `Instances` and `Backups` are. The default stays `%LocalAppData%\Borea`. |
| | Settings, App preferences, the index and image caches, logs and installed loaders stay in `%LocalAppData%\Borea`, because they are small and Borea reads its settings before it knows another folder. |
| | The setting is in `BoreaSettings`, so the App and the CLI see the same instances. The App changes it in the settings, and the CLI with `borea settings set library <dir>`. `borea settings show` prints it. |
| | A change moves the existing instances and backups. On the same drive Borea renames the folders; to another drive it copies them with progress and deletes the old files only after the copy is complete. |
| | Borea refuses the move while it runs the game for an instance or while a file is locked. When the move fails, the old folder stays in use, and Borea saves the setting only after the move succeeded. |
| | When the chosen folder already holds a Borea library and the current one has no instance, Borea uses it without a move, for example after a Windows reinstall. When both hold instances, Borea refuses to merge them. |
| | Borea refuses a folder inside the game folder, inside `My Games/Kitten Space Agency`, inside the current library, and a folder that it cannot write to. |

### Tests

- A move on one drive and a copy to another root, with the instances, backups and records intact afterwards.
- A failure during the copy leaves the old library in use and no partial copy.
- A locked file and a running launch refuse the move.
- An existing library in the chosen folder is used without a move, and two libraries with instances are refused.
- Each refused folder.
- The CLI command and `settings show` with `--json`.

### Not in scope

- Moving settings, caches, logs or loaders.
- The shared store of static mod files (#84), which will live in the library folder.

Maximilian-Nesslauer:
Closes #273

- Settings > General has a new "Library folder" row. It shows where Borea keeps instances and backups, with "Move library...", "Open folder", and "Move back to the default folder" when another folder is set. A hint and tooltips say that moving takes the instances and backups along and that "Open folder" moves nothing.
- The default stays `%LocalAppData%\Borea`. Settings, preferences, the caches, the logs and installed mod loaders always stay there.
- On the same drive Borea renames the `Instances` and `Backups` folders. To another drive it copies them with a progress line and a Cancel button, compares the file count and the sizes, saves the setting, and only then deletes the old files.
- A failed or cancelled copy deletes the partial copy, and the old library stays in use. Closing Borea during a copy cancels it the same way before the window closes.
- When the chosen folder already holds a Borea library and the current one has no instance, Borea uses it without a move, for example after a Windows reinstall. When both hold instances, Borea refuses and names both folders.
- Borea refuses a relative path, a file, a folder inside the game folder or inside `My Games/Kitten Space Agency`, a folder it cannot write to, the current library and a folder inside or around it, and a folder inside or around `%LocalAppData%\Borea` other than that folder itself. These checks follow links and junctions.
- Borea refuses the change while the game runs, while another Borea window or command runs, or while a file in the library is open in another program. While the change runs, the App refuses installs, updates, removals, launches, and changes to instances and saves, and while one of those runs, the App refuses the change.
- On the command line, `borea settings set library <dir>` runs the same change, and `borea settings set library --default` moves the library back. `borea settings show` prints the library folder.


averageksp (dismissed):

MrJeranimo (changes requested):
Everything looks good and works on my machine, even when I try to move the folder to a OneDrive folder it works. 

There is only one thing that needs to be added. The result of the change, and that it was tried, wasn't put in the output/activity log (Tasks area I think). It is a big enough change that it should show up there, so the user can see that they did change the folder location if they forget.

Maximilian-Nesslauer:
@MrJeranimo Done in 62724e1 and 6bd0b44: every library folder change now shows in the task history, with the target folder and the result (moved, refused, failed or stopped, with the reason). A toast says where the library is now or why the change failed, the task drawer shows the copy progress, and the Borea log records the change for the App and for `borea settings set library`.

averageksp (approved):

averageksp (approved):

MrJeranimo (approved):
Looks good to me, also works on my machine and the Tasks Log now shows the changes.

PR #316, opened 2026-09-18
Show the download size before an install, and compact the download counts

averageksp:
## Show the download size before an install, and compact the download counts

The index snapshot carries the archive size of every release, and Borea did not show it anywhere before the download started (MrJeranimo on Discord, 2026-09-18). Large download counts also took a lot of a row with their thousands separators (beik on Discord).

### Download size

- Discover rows show the archive size of the release Add installs, with a hard drive icon before the download count and the age. It follows the release channel like the age does.
- The mod page shows the same size next to the version in the header, and the Versions table gets a Size column between Compatibility and Published. The column hides at narrow widths like Published.
- The confirm button of an install says what the plan downloads: "Add (412.0 MB)" or "Install anyway (412.0 MB)", summed over every release in the plan, so dependencies count. Same for packs. Without a stated size, the button reads as before.
- A release without a size, which is what SpaceDock listings have, shows nothing.

### Download counts

- Rows and the mod page show 1.2k, 15k, 1.2M instead of 1,234, 15,300, 1,234,567. One decimal below 10, none above, rounded down so a mod never shows a figure it has not reached.
- The tooltip on the count has the exact number, "1,234 downloads".
- The decimal separator follows the language.

### How

- `MainViewModel.SizeText` is the one size formatter, taken out of `FormatGameDataSize`. Decimal units, so the number agrees with the download progress.
- `MainViewModel.CompactCount` formats the counts. `MainViewModel.PlanSizeText` sums `Download.SizeBytes` over the plan's operations.
- `DiscoverItem.DownloadSizeText` reads `LatestInChannel`, which now notifies. `VersionItem.DownloadSizeText` reads its release. `DiscoverItem.DownloadsExactText` is the tooltip.
- `ConfirmInstallText` on the Discover, version and pack rows comes from one method, and follows `Choices` as well as `InstallWarning`, because the choices path holds a plan without a warning.
- `Icon.HardDrive` from Phosphor, converted to the 24 grid of `Icons.axaml`.
- `ContentDownloadSize`, `ContentSizeHeader` and `ContentDownloadsExactFormat` in English and German.

### Tests

542 pass. `SizeTextTests` covers `CompactCount` at each boundary and `SizeText` in each unit. The Discover test checks the row against the formatter and the exact tooltip.

### Not in this PR

The unpacked size. The index has it, but two numbers on a row is noise. It could go into the tooltip later.


src/Borea.App/ViewModels/MainViewModel.Discover.cs
```diff
@@ -659,11 +669,13 @@ public sealed partial class DiscoverItem : ObservableObject, IInstallRow
 
     [ObservableProperty]
     [NotifyPropertyChangedFor(nameof(IsConfirmingInstall))]
+    [NotifyPropertyChangedFor(nameof(ConfirmInstallText))]
```
Maximilian-Nesslauer, line 681: [should fix] This notification fires before the plan is known.
`PlanAndExecuteAsync` sets `row.Choices` and only then calls `HoldPlan`, which sets `PendingPlan`.
`PendingPlan` is a plain property, and `InstallWarning` goes from null to null when the plan has no warning, so nothing reads `ConfirmInstallText` again and the button stays at "Add" without the size.
The pack path has the same order (`pack.Choices = choices; HoldPack(...)`).

Making `PendingPlan` an `[ObservableProperty]` with `[NotifyPropertyChangedFor(nameof(ConfirmInstallText))]` on `DiscoverItem`, `VersionItem` and `PackItem` fixes both paths, and the text then also follows a new plan after Confirm.
The fixtures in `InstallChoicesTests` have `sizeBytes: null`, so a test with a size there would pin it.
The live index has no release with `recommends`, `suggests` or `any_of` yet, so the App cannot show the case today.

src/Borea.App/ViewModels/MainViewModel.GameData.cs
```diff
@@ -83,6 +84,47 @@ internal string FormatGameDataSize(long bytes)
...
+    }
+
+    private static string Compact(double value, string unit)
+    {
+        var rounded = value < 10 ? Math.Floor(value * 10) / 10 : Math.Floor(value);
+        return rounded.ToString(value < 10 ? "0.#" : "0", CultureInfo.CurrentCulture) + unit;
+    }
+
+    /// <summary>What a plan downloads, summed over its operations, or null when no release states a size.</summary>
+    internal static string? PlanSizeText(InstallPlan? plan)
```
Maximilian-Nesslauer: [minor] `PlanSizeText` is about install plans and pulls `Borea.Core.Planning` into the game data partial.
It would fit better next to `ConfirmInstallText` in `MainViewModel.Install.cs`.

Maximilian-Nesslauer (changes requested):
Thanks, this is useful :)

[should fix] The instance page shows no sizes yet: the mod rows have none, and the instance has no total. I would add both in this pull request. For installed content the useful number is the space on disk and not the download size, so I would measure the folders the way `FileGameDataReader` already does for the game data rows. That also covers manual installs and files that a mod writes itself, which `install_size` from the index would miss. The mod rows can use the `SizeText` of this pull request, and the total of the whole instance folder could sit in the instance header.

[question] The size only shows when there is a confirmation step. A plan without a warning or a choice installs at once, so the total never shows before the download. With the live index, "+" on KSP-Redux-OPM shows 480.7 MB on the row and then also downloads KSP-Redux, about 2.3 GB in total, without asking. Should an install ask when its dependencies add to the download, or should the row show the total of the plan?

averageksp:
> Thanks, this is useful :)
> 
> [should fix] The instance page shows no sizes yet: the mod rows have none, and the instance has no total. I would add both in this pull request. For installed content the useful number is the space on disk and not the download size, so I would measure the folders the way `FileGameDataReader` already does for the game data rows. That also covers manual installs and files that a mod writes itself, which `install_size` from the index would miss. The mod rows can use the `SizeText` of this pull request, and the total of the whole instance folder could sit in the instance header.
> 
> [question] The size only shows when there is a confirmation step. A plan without a warning or a choice installs at once, so the total never shows before the download. With the live index, "+" on KSP-Redux-OPM shows 480.7 MB on the row and then also downloads KSP-Redux, about 2.3 GB in total, without asking. Should an install ask when its dependencies add to the download, or should the row show the total of the plan?

On the question: I'd have the install ask when the plan pulls in more than the mod itself. "+" on KSP-Redux-OPM downloading 2.3 GB without a word isn't what anyone expects, and the row can't show the total because the plan only exists after the click. A line like "Also adds KSP-Redux 1.3.2" with the button "Add (2.3 GB)" would do it, using the same confirmation the warnings use. A mod with a small library dependency would ask once too, which I think is fine. Separate PR though, since it changes when installs ask.

tests/Borea.App.Tests/ViewModels/InstallChoicesTests.cs
```diff
@@ -40,6 +40,22 @@ public async Task Install_Recommendations_InstallTheKeptOneAndSkipTheDeselectedO
...
+        using var harness = await CreateWithGameAsync();
+        var release = Release(OwnId, dependencies: [Recommends("kept")], sizeBytes: 30_000_000);
+        harness.SpaceDock.Releases.AddRange([release, Release("kept", sizeBytes: 8_000_000)]);
+        await ActivateInstanceAsync(harness);
+        var row = new VersionItem(harness.ViewModel, release);
+
+        await row.InstallCommand.ExecuteAsync(null);
+
+        Assert.True(row.IsConfirmingInstall);
+        Assert.Equal($"{harness.Localization.ContentAdd} ({MainViewModel.SizeText(38_000_000)})", row.ConfirmInstallText);
```
Maximilian-Nesslauer: [should fix] This passes without the fix too.
The getter already returned the size before, only the notification came too early, and this line reads the getter directly.
I turned `PendingPlan` back into a plain property on the three rows and this test still passed.

Recording what a binding sees pins it:

```csharp
var row = new VersionItem(harness.ViewModel, release);
string? shown = null;
row.PropertyChanged += (_, e) =>
{
    if (e.PropertyName == nameof(VersionItem.ConfirmInstallText))
        shown = row.ConfirmInstallText;
};

await row.InstallCommand.ExecuteAsync(null);

Assert.True(row.IsConfirmingInstall);
Assert.Equal($"{harness.Localization.ContentAdd} ({MainViewModel.SizeText(38_000_000)})", shown);
```

With the fix this passes, and without it it fails with `Expected: "Add (38,0 MB)"`, `Actual: "Add"`.

src/Borea.Storage/Instances/FileInstanceSizeReader.cs
```diff
@@ -0,0 +1,64 @@
...
+    {
+        var mods = new Dictionary<string, long>(ModIds.Comparer);
+        var modsFolder = new DirectoryInfo(_pathProvider.GetInstanceModsFolder(instanceId));
+        if (modsFolder.Exists)
+        {
+            foreach (var folder in modsFolder.EnumerateDirectories())
+                mods[folder.Name] = Size(folder, cancellationToken);
+        }
+
+        return new InstanceSizes(Size(new DirectoryInfo(_pathProvider.GetInstanceRoot(instanceId)), cancellationToken), mods);
```
Maximilian-Nesslauer: [minor] The mod folders are walked twice: once each above, and again inside this walk of the instance root.
The total could be the sum of `mods` plus a walk of the root that skips the `mods` folder, so a big instance is read once when its page opens.

Maximilian-Nesslauer (changes requested):

Maximilian-Nesslauer (approved):
All points are fixed, thanks :)

PR #335, opened 2026-09-18
Log when an instance is created, changed or deleted

### LINKED ISSUE #319: Log when an instance is created, changed or deleted
### what

On 2026-09-18 three instances disappeared during an App test, and the Borea log could not say what deleted them.
Today only the library folder move of #273 writes a log line about instances.

| Today | Should be |
|---|---|
| Creating, renaming, duplicating, deleting and importing an instance (game profile, modlist) writes no log line. | Each of these writes one line with the instance name and id. |
| Activating and deactivating an instance writes no log line. | Both write one line. |
| | The CLI writes the same lines as the App, because both use the same code. |

Maximilian-Nesslauer:
Closes #319

- The Borea log now has one line when an instance is created, renamed, deleted, activated or deactivated.
- Each line names the instance and its id, so you can find all lines of one instance with a search.
- A created line says where the instance came from: a new empty instance, a duplicate, a modlist import, the game profile import, or a mod pack.
- A delete writes one line when it starts, with the instance folder, and one line with the result. When the delete fails part way, the second line says so, and says whether Borea still lists the instance.
- A delete of an instance that does not exist writes "not found, nothing deleted".
- A taken name or an unknown instance writes a short "refused" line without a stack trace. Other failures write the error with its details.
- The App and the CLI write the same lines to the same daily log, because both use the same instance repository.

Example (real output from an App test, time stamps removed, ids shortened):

```
Instance "test123" (2246795e...) created as a new instance.
Instance "test123" (2246795e...) activated.
Instance "test123" (2246795e...) renamed to "test1234".
Instance "test1234" (2246795e...) deactivated.
Instance "test1234 (copy)" (ae507465...) created as a duplicate.
```


MrJeranimo (approved):
Logging works beautifully. Also makes sense why this doesn't show up in the Tasks area as there may be many Instance operations. That would clutter up the important logs.

PR #336, opened 2026-09-18
Start a loader by its per-platform launch entry (RFC 0067)

### LINKED ISSUE #321: Start a loader by its per-platform launch entry (RFC 0067)
### what

[RFC 0067](https://github.com/KSAModding/content-manager-design/pull/67) lets a loader listing say how the loader starts on one platform, in `[provides.platform.<os>]`.
Borea still guesses (#225): on Linux and macOS, when `launch` ends in `.exe` and a `.dll` with the same name is beside it, Borea starts `dotnet` with the `.dll`.

| Today | Should be |
|---|---|
| The guess of #225. | Borea reads the entry of its own platform from the live listing and starts its `launch`, or the named `runtime` with the absolute entry file as the first argument. |
| | Without an entry for its platform, Borea starts `[provides].launch`, as RFC 0035 says. |
| | An unknown runtime or key, a runtime that is not on the PATH, or an entry file that is not in the installed loader stops the launch with the reason. There is no fallback to `[provides].launch`. |
| | The handover flag and the instance root of RFC 0049 and the launch arguments of #269 come after the entry file. |
| | The guess of #225 is removed. |

### Order

This merges after KSAModding/content-index#83 and after the StarMap listing carries its `linux` and `macos` entries, so StarMap keeps starting on Linux.

Maximilian-Nesslauer:
Closes #321

- A loader listing can now say how the loader starts on one platform, in `[provides.platform.windows]`, `[provides.platform.linux]` or `[provides.platform.macos]` (RFC 0067).
- Borea reads the entry of its own platform from the live listing at every launch, the same way it reads `[provides.instance]`.
- With an entry for its platform, Borea starts the entry's `launch`. Without one, it starts `[provides].launch` as before.
- With `runtime = "dotnet"`, Borea finds `dotnet` on the PATH, then in `DOTNET_ROOT`, then in the usual install folders, and starts it with the absolute path of the entry file as the first argument.
- The working directory stays the loader folder. The instance flag and folder of RFC 0049 and the saved and one-time launch arguments come after the entry file.
- Borea starts nothing and says why when the entry of its platform has an unknown runtime or key, when `dotnet` is not found, or when the entry file is not in the installed loader. It does not fall back to `[provides].launch`.
- Unknown names or keys in the entries of other platforms change nothing, and a broken entry only rejects that one loader listing.
- The old guess is removed: Borea no longer starts `dotnet` with a `.dll` it finds beside a `.exe` on Linux and macOS. StarMap needs its `linux` and `macos` entries in the listing for that.
- The App shows the new reasons in English and German. The CLI prints them and exits with 1.

Example listing entry:

```toml
[provides.platform.linux]
launch = "StarMap.dll"
runtime = "dotnet"
```


MrJeranimo (approved):
LGTM

PR #337, opened 2026-09-18
Show launch and instance errors without pushing the layout

### LINKED ISSUE #284: Show launch and instance errors without pushing the layout
### what

Errors on the instance page are inserted into the middle of the layout and push the content around (beik on Discord, 2026-09-17).

| Today | Should be |
|---|---|
| A loader that stops right after Play (#220) shows a block with the reason, Disable, Show details and Open launch log inside the page. | A modal shows the reason, "Disable <mod>", the details and "Open launch log". The page shows one status line at most. |
| Short results and errors of instance actions (rename, remove, game data, saves) appear as text in the page. | They show as toasts (#282), and their details go to the Log tab or the Tasks page. |
| | The number of elements that appear in the page layout because of an error stays small. |

### Tests

- A crashed start opens the modal with its actions, and the page keeps one status line.
- An action error shows a toast and no text block in the page.

Depends on #282.

Maximilian-Nesslauer:
Closes #284

- When the mod loader stops right after Play, a modal now opens. It says why the loader stopped and offers "Disable <mod>" when a mod is named, "Show details" and "Open launch log".
- Home, the Library and the instance page keep one status line for the launch. Its "Show details" link opens the modal again after you closed it.
- Short results and errors of instance actions now show as toasts instead of text in the page. This covers activate, deactivate, delete, duplicate, open folder, enable and disable a mod, remove a mod, export, copy and import a modlist, the saves and vehicles actions, the game data and log actions, and the manual installs actions.
- An error toast writes its full text to the Borea log. Its "Show details" opens a small modal with the text and "Open Borea log".
- Errors that belong to a field stay at the field, for example a taken name in the name modal.
- When you close the name modal or the launch arguments modal before the save finishes and the save fails, the error shows as a toast.
- A tab that cannot load (game data, log, manual installs, a saves section) still says so in the tab, because that is a state and not a result.


MrJeranimo (approved):
Love the Toasts, they look so much better than the plain text.

PR #338, opened 2026-09-18
Pause and resume a download

### LINKED ISSUE #318: Pause and resume a download
### what

Stop (#271) cancels a download and deletes the part that was downloaded.
A player on a slow line who has to interrupt a large download, such as KSP-Redux with 1.7 GiB, starts again from zero (averageksp on Discord, 2026-09-16).

| Today | Should be |
|---|---|
| A running download can only be stopped. | Next to Stop, Pause keeps the downloaded part, and Resume continues from there. |
| | Resume asks the host only for the rest of the file. When the host does not support that, or the file changed on the host, the download starts again from zero. |
| | The SHA-256 check of the whole file stays, so a resumed download installs only the bytes that the index stamped. |
| | Pause and Resume are on every row that shows Stop and in the Tasks drawer. A paused task shows as paused. |

### Not in scope

- Keeping a paused download when Borea closes. Closing Borea stops it, as it does today.
- Pause in the CLI.

### Tests

- A paused and resumed download requests the rest with a range, and the result passes the digest check.
- A host that answers a range request with the whole file restarts the download from zero.
- A changed file on the host restarts from zero and never mixes old and new bytes.
- Stop on a paused download deletes the part.

Maximilian-Nesslauer:
Closes #318

- Next to Stop, every running download now has Pause. This is on the Discover rows, the mod page, the pack page and the Modpacks rows, the updates of one mod and Update all, the mod loader install, and the Tasks drawer.
- Pause keeps what was downloaded. Resume asks the host only for the rest of the file, so a large mod on a slow line does not start again from zero.
- When the host cannot send only the rest, or the file on the host changed, the download starts again from zero. Old and new bytes are never mixed.
- The SHA-256 check of the whole file stays, so a resumed download installs only the bytes that the index stamped.
- When the host fails after Resume, the next mirror starts from zero, as before.
- Pause works only while a mod downloads. While the mod unpacks and installs, the button is off.
- A paused task shows "Paused" in the Tasks drawer, the row shows how much was downloaded, and the progress bar in the navigation stays at that value.
- Stop on a paused download stops it and deletes the part. Closing Borea also deletes it, so a pause does not survive a restart.
- The CLI does not change.


MrJeranimo (approved):
LGTM. This was a very nice change and it does work wonderfully.

PR #350, opened 2026-09-19
Add the Borea landing page on GitHub Pages

### LINKED ISSUE #348: Add a simple landing page for Borea
It would be helpful for players and mod authors to have a simple site with download instructions and a guide on how to list a mod in the app.

Here we have the plan for a first version

[borea-landing-page-plan.md](https://github.com/user-attachments/files/32413257/borea-landing-page-plan.md)

Maximilian-Nesslauer:
Closes #348

- A one-page site with the app icon, the name, one line and one download button, then a screenshot of the Discover page and three short sections: About, For mod authors and Get involved.
- The button detects the OS and downloads the App archive for Windows, Linux or macOS from the latest stable release. Phones see "Available for Windows, Linux and macOS" in place of the button.
- The page is dark only and uses the tokens and IBM Plex fonts of the app. It has no framework, no analytics and no cookies.
- It is hosted at https://ksamodding.github.io/Borea/ and the `pages` workflow deploys `site/` on every push to `main` that changes it.
- Until the first stable release exists, the page shows the static "Download" fallback to the releases page, because `releases/latest` skips pre-releases.

For @renancamm, the places where the site differs from the plan:

- The "content index" link and the "MIT licensed" footer link keep a faint underline, because their color alone is too close to the text around them for WCAG 1.4.1.
- Plex Sans Regular is IBM's Latin1 subset plus the arrow of "List your mod", so the site still loads three font files.
- The screenshot also has a 760 px WebP for phones, and the icon is served as WebP with the PNG as fallback.
- The screenshot is lazy-loaded as the plan says, but on phones its top is the largest element in the first view, so the lazy load delays LCP (mobile Performance is 96). Eager loading would give more margin.
- The favicon is `assets/borea.ico` as it is (173 KB, 9 sizes). An `.ico` with only 16, 32 and 48 px would make the first load much smaller.
- The section headings use the heading style of the App's side panel ("Category", "Game version"): Plex Sans SemiBold in white at 18 px, instead of the grey Mono label, because the Mono labels did not read as headings below the body text.
- The sections are 64 px apart (48 px on phones) instead of 96 px, with 12 px from a heading to its text, and the footer rule starts 48 px below the last section.
- The sections and the footer share the 880 px column of the screenshot instead of a 640 px column, so both edges line up with it. From 760 px on, each section heading sits in a 200 px column to the left of its text, which keeps the text about 640 px wide; on narrower screens the heading stays above the text.
- The hero has fixed padding (120 px above, 80 px below; 72 and 56 on phones) instead of `min-height: 88vh`, so a tall window shows no empty gap before the screenshot, and the screenshot still shows at the bottom of the first screen.


averageksp (approved):
Looks beautiful

PR #396, opened 2026-09-19
Restore and clean up backups in the App

### LINKED ISSUE #368: Restore and clean up backups in the App
### what

Borea moves replaced and deleted saves, vehicles and mod folders into the Backups folder of the library (#215, #273), so nothing is lost by mistake.
The App cannot bring a backup back, and nothing ever removes old backups, so the folder only grows.

| Today | Should be |
|---|---|
| A player who wants a backup back has to find it in the file explorer and copy it by hand. | The instance page lists its backups, newest first, with what each one was and when Borea made it, and "Restore" puts it back. When the target exists, Borea asks first and backs that one up too. |
| Old backups stay forever. | A backup can be deleted in the App, and a General setting deletes backups older than a chosen age, off by default. |
| | The CLI gets the same as verbs of `borea instance`. |

Maximilian-Nesslauer:
Closes #368

- The Game data tab of an instance has a new Backups section. It lists the backups of the saves and vehicles of that instance, newest first. Each row shows the name, what happened (backed up, deleted or replaced), when, and the size.
- Restore puts a backup back into the saves or Vehicles folder it came from. When a save or vehicle with that name is there, Borea asks first and moves the current one into the backups too, so nothing is lost.
- A zip from "Back up" stays in the backups after a restore. A deleted or replaced folder leaves the backups when it goes back.
- Delete removes a backup for good. Borea asks first.
- When Borea cannot tell where a backup came from, for example a folder put there by hand, the row says so and offers only Delete.
- Restore waits until the game of that instance is closed. Delete also works while the game runs, because the game does not read the backups.
- Restore and Delete show in the task history and end with a message.
- Settings, General, has "Delete backups older than": Off (the default), 30, 90 or 365 days. Borea applies it to the backups of every instance at each start and when the setting changes, and writes the count to the log.
- Each new backup gets a small file next to it, `<backup>.borea.toml`, that says which folder it was, when Borea made it and why. For older backups Borea reads this from the name, and the time from the folder when the name has none.
- Only saves and vehicles have backups. Borea does not keep mod folders in the Backups folder.
- The CLI gets `borea instance backups <instance>`, `borea instance restore-backup <instance> <backup> [--replace]` and `borea instance delete-backup <instance> <backup>`, each with `--json`. A backup is named by its id from the list, or by its file name when only one backup has it.


MrJeranimo (approved):
Backups work nicely

PR #398, opened 2026-09-19
Show the dependencies on the content page

### LINKED ISSUE #392: Show Dependencies when in Content/Mod view
Currently in the Content information pages, when you click into a mod/modpack/modloader's information page, it doesn't show what the dependencies and their types are. The only way to see that is to try and install the mod, or through the description of the mod.

Image

This should most likely have a `Dependencies` tab between `Description` and `Versions`. The dependencies tab should show a list of dependencies, sorted into Required, Any Of, Conflicting, Recommended, Optional. This will show the list of dependencies in order of importance.

Maximilian-Nesslauer:
Closes #392

- The content page has a new Dependencies tab between Description and Versions.
- The tab shows the dependencies of the release in the page header. That is the release that Add installs in your release channel. When you change the channel, the tab changes too.
- When the release needs a mod loader, the loader comes first, for example "Needs StarMap 0.4.5 or newer".
- The dependencies are in groups, in this order: Required, Any of, Conflicts, Recommended, Suggested, Optional. A group shows only when it has entries. A short line under each group says what it means for you.
- In "Any of", each entry lists the mods you can choose from.
- Each entry shows the name and icon from the index, and a click opens its page. A mod that is not in the index shows its id and "Not in the index".
- Each entry shows its version bounds, for example "1.2.0 or newer" or "1.2.0 to 2.0.0".
- Each entry shows if the mod is installed in the instance that Add installs into, with its version. An installed mod under Conflicts is shown in red.
- A release without dependencies says "This release needs no other mods."
- A mod loader page shows the tab only when its release has dependencies.
- When your release channel has no release of a mod, the page header no longer keeps the release of the previous channel.
- A name that cannot be opened, such as a mod that is not in the index, no longer sits on a grey box, and a click on a tile no longer flashes in its rounded corners.


MrJeranimo (approved):
LGTM

Issue #312, opened 2026-09-18
Running Borea on macOS fails


Kiwi-Shark:
I've attempted to run Borea on my M4 MBA and each attempt fails to load the app.

### From building:
I can build Borea successfully via `dotnet build` in `Borea`.

I then `cd` into `src/Borea.App/bin/Debug/net10.0` and run `dotnet Borea.App.dll`, which gives the error message '"Borea.App.app" is damaged and can't be opened. You should move it to the Trash.'

Checking the console I see `Unhandled exception. System.TypeInitializationException: The type initializer for 'SkiaSharp.SKImageInfo' threw an exception.` followed by a pretty verbose output. `dlopen(libSkiaSharp.dylib, 0x0001): tried: 'libSkiaSharp.dylib' (no such file)` gets repeated a handful of times. It all ends with:

```
dlopen(liblibSkiaSharp, 0x0001): tried: 'liblibSkiaSharp' (no such file), '/System/Volumes/Preboot/Cryptexes/OSliblibSkiaSharp' (no such file), '/usr/lib/liblibSkiaSharp' (no such file, not in dyld cache), 'liblibSkiaSharp' (no such file)

   at SkiaSharp.SkiaApi.sk_colortype_get_default_8888()
   at SkiaSharp.SkiaApi.sk_colortype_get_default_8888()
   at SkiaSharp.SKImageInfo..cctor()
   --- End of inner exception stack trace ---
   at Avalonia.Skia.PlatformRenderInterface..ctor(Nullable`1 maxResourceBytes, Nullable`1 useStencilBuffers)
   at Avalonia.Skia.SkiaPlatform.Initialize(SkiaOptions options)
   at Avalonia.SkiaApplicationExtensions.<>c.<UseSkia>b__0_0()
   at Avalonia.AppBuilder.SetupUnsafe()
   at Avalonia.AppBuilder.Setup()
   at Avalonia.AppBuilder.SetupWithLifetime(IApplicationLifetime lifetime)
   at Avalonia.ClassicDesktopStyleApplicationLifetimeExtensions.StartWithClassicDesktopLifetime(AppBuilder builder, String[] args, Action`1 lifetimeBuilder)
   at Borea.App.Program.Main(String[] args) in /Users/edbiv/Documents/GitHub/Borea/src/Borea.App/Program.cs:line 22
zsh: abort      dotnet Borea.App.dll
```
### From release
[...]

PR #411, opened 2026-09-20
Fix MacOS `AccessViolationException` due to compressing the release to a single file.

MrJeranimo:
Change the `EnableCompressionInSingleFile` option to `false` for `osx-arm64` (MacOS) build versions. This was causing an `AccessViolationException` on MacOS that was causing the app to fail 90% of the time. 

This was brought to my attention by KiwiShark [Issue Link](https://github.com/KSAModding/Borea/issues/312#issuecomment-5746908489).

According to Claude, the issue was the `EnableCompressionInSingleFile` option was causing the `HTTPConnectionPool` to write to protected memory during startup, thus causing a crash. This is a known .NET 10.0 bug but has not been fixed yet. The fix was to change `EnableCompressionInSingleFile` on `osx-arm64` builds. 

Claude's words below,
```
Read memory, searched the web, read a page
That crash trace is a strong match for a known .NET host bug on macOS Apple Silicon, not a bug in Borea's networking code itself.

What's happening

The exception is thrown deep inside System.Net.Http.HttpConnectionPool, triggered from Borea.Network.Announcements.AnnouncementFetcher — i.e. Borea is making its startup "check for announcements" HTTP call, and the runtime corrupts memory while setting up that connection. The giveaway is the exception type itself: AccessViolationException inside otherwise-safe managed code means something below the managed layer (the .NET host/runtime) wrote to memory it shouldn't have — this isn't something application-level C# code can normally cause.

There's an open .NET runtime issue describing exactly this pattern: on Apple Silicon, self-contained single-file apps published with compression enabled crash with AccessViolationException a large percentage of the time, especially when concurrent async work (like a network call or spawning a process) happens early during startup. The reporter notes the exact stack trace varies between runs — sometimes it's in process-spawning code, sometimes elsewhere — which is typical of memory corruption rather than a single deterministic bug in one code path. That matches this Borea crash landing specifically in the HTTP stack instead. 
[GitHub](https://github.com/dotnet/runtime/issues/123324)
[GitHub](https://github.com/dotnet/runtime/issues/123324)

Key details from that issue:

Reproduces on .NET 7, 8, 9, and 10 on osx-arm64, with failure rates in the 90%+ range for the minimal repro.
The known trigger is PublishSingleFile + EnableCompressionInSingleFile=true in the macOS arm64 publish profile.
It's tracked against a .NET 11 milestone; a fix PR (
#127355) exists but hasn't shipped in a released runtime yet.
What to check/do

Look at the macOS arm64 publish profile/csproj for Borea's release build — if EnableCompressionInSingleFile is true, that's almost certainly the trigger.
The documented workaround is to set EnableCompressionInSingleFile=false for the osx-arm64 publish target. This trades a somewhat larger binary for eliminating the crash.
A Task.Delay at startup before concurrent async work reduces (but doesn't eliminate) the failure rate — not a real fix, just evidence for diagnosis if you want to confirm the theory.
Since this is a host bug rather than an app bug, no amount of changing AnnouncementFetcher's own code will reliably fix it — the fix has to be at the publish-settings level (or eventually a runtime update).
```


averageksp (approved):

Maximilian-Nesslauer (approved):
LGTM  👍 

Two small things: the issue dotnet/runtime#123324  is still open, so "tracked for .NET 11" is more exact than "fixed for", and `RuntimeIdentifier.StartsWith('osx')` would also hold for a future macOS runtime. Neither blocks the merge.

What is still missing is one start on a real Mac.

PR #424, opened 2026-09-20
Catch a crashing macOS build in the release checks

### LINKED ISSUE #423: Catch a crashing macOS build in the release checks
### what

#411 fixed a crash that made the macOS build fail on about nine of ten starts (#312), and the release workflow reported everything green.
`verify archives (macos-arm64)` runs `borea --help`, `borea --version`, `borea no-such-command` and `borea settings show --json`.
None of them makes a network request, and the crash of dotnet/runtime#123324 needs early concurrent work such as an HTTP call, so the check could not see it.

| Today | Should be |
|---|---|
| The archive check runs only commands that stay local. | The check also runs a command that makes a request, for example `borea index refresh`, so a start that crashes in the HTTP stack fails the run. |
| | On macOS it runs that command three times in a row, because the crash does not happen on every start. |
| | A host that cannot be reached is not a failure. Only a crash of the program is, so a release does not go red when GitHub Pages is slow. |
| Nothing states which build settings a platform must have. | The release scripts test states that `EnableCompressionInSingleFile` is off for every macOS runtime and on for the others, so the fix of #411 cannot be undone by accident. |

The second line is the deterministic part and the first line is the behavior part. Both are needed, because the property can be right while the runtime still crashes, and the other way round.

### Tests

- The release scripts test fails when `EnableCompressionInSingleFile` is true for a macOS runtime.
- A manual run of the release workflow shows the new command in the log of all three platforms.

Maximilian-Nesslauer:
Closes #423

- The release checks now start `borea index refresh` from the App archive on every platform, and three times on macOS. A start that makes a request is what showed the crash of dotnet/runtime#123324, which #411 fixed. The checks before it ran only commands that stay local, so the release looked green.
- A host that does not answer does not fail the run. Exit code 1 means that the operation failed, which is what an unreachable host gives, so only a higher code counts as a crash.
- A new test in the release scripts job reads the publish settings from MSBuild and states them: every runtime publishes one single file, a macOS runtime never compresses it, and every other runtime does. It reads the runtime list from the projects, so a runtime that is added later is covered too.


averageksp (approved):

PlazmaBoltz (approved):

MrJeranimo (approved):
Hopefully this can catch more errors in the future.

